PR #15, opened 2016-08-31
Done some changes to Fredboat.java

RealKC:
Removed some useless stuff. Added an alias for ;;s that's easier to remember IMO.


Frederikam:
:+1:

PR #16, opened 2016-08-31
Ram Command

RealKC:
;;ram is now a thing. xD


src/main/java/fredboat/FredBoat.java
```diff
@@ -247,6 +247,9 @@ public static void init() {
         CommandRegistry.registerCommand(0x11, "internetspeed", new RemoteFileCommand("http://www.speedtest.net/result/5529046933.png"));
         CommandRegistry.registerCommand(0x11, "hug", new RemoteFileCommand("http://i.imgur.com/U2l7mnr.gif"));
         
+        //Other Anime Discord memes//
+        CommandRegistry.registerCommand(0x11, "ram", new RemoteFileCommand("https://cdn.discordapp.com/attachments/132490115137142784/220508412662775810/ram.jpg"));
```
Frederikam: Files stored on Discord's CDN are deleted after a while of not being downloaded. Please use something like imgur instead.

RealKC:
Actually, wait a little before merging this.

RealKC:
Should be mergeable now.

PR #30, opened 2016-10-16
Update some things

Noahkiq:
Removed all mentions of the 20 song limit and added link to patreon where it previously said a patreon was coming soon:tm:


Frederikam:
Looks great!

PR #32, opened 2016-11-21
Added ;;cooldog command and some aliases

RealKC:
;;cooldog has the aliases ;;dogmeme and ;;dog
added ;;1ee7 alias to ;;leet

### DIFF
--- a/src/main/java/fredboat/command/fun/DogCommand.java
+++ b/src/main/java/fredboat/command/fun/DogCommand.java
@@ -31,9 +31,10 @@ public class DogCommand extends TextCommand {
 + "Spread him to take over Discord!(Don't spam him though!)";
 //dog created by @Ace#8925
 
-        public TextCommand() {
+        public DogCommand() {
                 super(DOG);
         }
 }
 
 //gib pat to me now
+//BTW future RealKC if you can't figure out what this code does, past you is disappointed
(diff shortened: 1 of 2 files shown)


Frederikam:
Please create a separate class for this. Preferably overriding `TextCommand`. This looks like a mess to have in the main class.

RealKC:
OK, I assume it will have to import the same things  TextCommand.java does.

Frederikam:
Mostly, but it needs to `extends TextCommand`. I recommend getting an IDE to handle the import statements.

src/main/java/fredboat/command/fun/DogCommand.java
```diff
@@ -0,0 +1,39 @@
...
++ "┈┈╲╲┈┈┈┈┈▏┈▏┈▔▔▔▆ ------- BORF!\n"
++ "┈┈┈╲▔▔▔▔▔╲╱┈╰┳┳┳╯ \n"
++ "╱╲╱╲▏┈┈┈┈┈┈▕▔╰━╯ \n"
++ "▔╲╲╱╱▔╱▔▔╲╲╲╲┈┈┈ \n"
++ "┈┈╲╱╲╱┈┈┈┈╲╲▂╲▂┈ \n"
++ "┈┈┈┈┈┈┈┈┈┈┈╲╱╲╱┈\n"
++ "Spread him to take over Discord!(Don't spam him though!)";
+//dog created by @Ace#8925
+
+        public TextCommand() {
```
Frederikam: Should be `DogCommand` instead of `TextCommand`

RealKC:
Idk why Travis has problems.

PR #45, opened 2017-01-08
main>personalfork

dotdog20:


Frederikam:
Never push your bot token, bot secret or Google API key on a public git repository. These credentials can be seen by anyone and can't be easily removed from git history.

Issue #28, opened 2016-10-03
Create ;;serverinfo command


Frederikam:
Displays info about a server

takoz53:
Should be displayed as following:

```C#
"Welcome to" + getServername() + ".\n
This Server currently has " + getUserCount() + "User(s).\n
Right now there are / is " + getOnlineUsers()  + "online User(s) in this Server.\n
There are " + getBotCount() + "Bot(s) available.\n
Have fun in the Server!"
```

Would be C# here tho; Convert to Java xD

Frederikam:
FredBoat shouldn't be giving out welcomes unless configured to do so, and I have no plans to make that configurable. I just want to bluntly show stats about a server. Take a look at the ;;stats command, which shows global stats.

Issue #25, opened 2016-10-03
Create ;;invite command


Frederikam:
Prints invite links to the bots, with basic instructions

goti921:
I would like to contribute.
Any more details?

PR #59, opened 2017-01-19
Added 2 cmd. Done #25,#28

knight-ryu12:


knight-ryu12:
Thank god Fred :+1:

FredBoat/src/main/java/fredboat/commandmeta/abs/ICommand.java
```diff
@@ -25,13 +25,14 @@
...
+import com.mashape.unirest.http.exceptions.UnirestException;
 import net.dv8tion.jda.core.entities.Guild;
 import net.dv8tion.jda.core.entities.Member;
 import net.dv8tion.jda.core.entities.Message;
 import net.dv8tion.jda.core.entities.TextChannel;
 
 public interface ICommand {
 
-    public abstract void onInvoke(Guild guild, TextChannel channel, Member invoker, Message message, String[] args);
+    public abstract void onInvoke(Guild guild, TextChannel channel, Member invoker, Message message, String[] args) throws UnirestException;
```
Frederikam: This change affects all commands. Please handle this with a try/catch in the ;;invite command

FredBoat/src/main/java/fredboat/command/util/InviteCommand.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+public class InviteCommand extends Command {
+    @Override
+    public void onInvoke(Guild guild, TextChannel channel, Member invoker, Message message, String[] args)  {
+        try {
+            String str = "https://discordapp.com/oauth2/authorize?&client_id=" + DiscordUtil.getApplicationInfo(invoker.getJDA().getToken().substring(4)).getString("id") + "&scope=bot";
+            String send = MessageFormat.format(I18n.get(guild).getString("invite"),DiscordUtil.getApplicationInfo(message.getJDA().getToken().substring(4)).getString("name"));
+            channel.sendMessage(send + "\n" + str).queue();
+        } catch (UnirestException e) {
+            e.printStackTrace();
```
Frederikam: This is not a proper way to log an Exception nor does it alert the user that something has gone wrong. Please use this:
```java
throw new RuntimeException(e);
```

Frederikam:
Great, now you just need to add the strings to the translation project:
https://github.com/Frederikam/FredBoat-i13n/blob/l10n_master/en_US.properties

Issue #76, opened 2017-02-11
[Future Project] adding comment on config/creds.json


knight-ryu12:

Frederikam:
You can't really add comments to a json file. You can add notes in the form of string values, but that's an ugly hack.

Alternatively you can parse the json files while ignoring comments, but this is something org.json is incapable of.

knight-ryu12:
Yeah, that's what I think of,
Kinda need better document about config/creds.json

napstr:
Why not use a different file format for the config, one which allows comments?

Frederikam:
I have actually been thinking of using something else like javascript, but I have much bigger priorities atm.

PR #86, opened 2017-03-06
Use YAML for the config files

napstr:
Substituted the JSON parser with a YAML one. Since YAML is a superset of JSON we don't even need to touch the existing config files = nothing should break anywhere, and can now add comments in YAML style (using #).

#76 can be easily done with this


Frederikam:
Looking at the source this looks great! I will test this locally once I'm not procrastinating from my homework.

napstr:
If you think it's viable I can write code that would use new .yaml files and convert existing json files into yaml ones for a smooth transition for everyone. While not functionally necessary, yaml files just look better than json ones.
Here's how credentials.yaml.example could look:

```
---
malPassword: 
clientToken: 
token:
  beta: 
  production: 
  music:

  # add your discord bot token below and remove the #
  # find the token of your bot on https://discordapp.com/developers/applications/me

  #patron: YourTokenHere

# add your google API keys in the brackets below, separated by commas if more than one, uncomment by removing the #
# how to find them: https://developers.google.com/youtube/registering_an_application

#googleServerKeys: [YourYoutubeAPIKey]
#googleServerKeys: [Key1, Key2]

cbUser:
cbKey: 
mashapeKey: 
```

Frederikam:
That'd be great, especially if you also take a look at the `gh-pages` branch.

FredBoat/src/main/java/fredboat/Config.java
```diff
@@ -67,60 +66,60 @@
...
-                jdbcUrl = creds.getString("jdbcUrl");
-            } else {
-                jdbcUrl = "";
+            mashapeKey = creds.get("mashapeKey") != null ? (String) creds.get("mashapeKey") : "";
+            malPassword = creds.get("malPassword") != null ? (String) creds.get("malPassword") : "";
+            carbonKey = creds.get("carbonKey") != null ? (String) creds.get("carbonKey") : "";
+            cbUser = creds.get("cbUser") != null ? (String) creds.get("cbUser") : "";
+            cbKey = creds.get("cbKey") != null ? (String) creds.get("cbKey") : "";
+            Map<String, String> token = (Map) creds.get("token");
+            botToken = token != null ? (token.get(distribution.getId()) != null ? token.get(distribution.getId()) : "") : "";
```
Frederikam: These ternary statements look unnecessary cluttery. Why not just use `Map#getOrDefault()`?

napstr: Good point. Haven't paid much attention to 1.8 stuff.

Frederikam:
I really like this change, this should make it much easier for people to start selfhosting the bot.

Please use the `development` branch next time. Not much of a biggie now as the branches are mostly even.

Thanks!

PR #88, opened 2017-03-08
files for running the boat as a systemd service

napstr:
want to share a chill way to run the boat for selfhosters

additional documentation has been added to the selfhosting page too

doing this PR against master as it isn't touching any code, just adding sample files


run.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+#README
+# your system needs to be using systemd, for example Ubuntu 15.04 and later; tested on 16.04
```
Frederikam: This is a bit misleading. This shell file does not necessarily need systemd to be useful. Though of course, the instructions in the comments are for systemd.

Issue #42, opened 2017-01-05
A third repeat mode that moves tracks to the tail instead of restarting


Frederikam:
If I were to implement this, each player would have one of 3 repeat modes:

1. Ended tracks are deleted (off)
2. Ended tracks are restarted immediately (current ;;repeat)
3. Ended tracks are shuffled back into the queue or put at the the tail, depending on shuffle mode (proposed idea)

I'd like some input on this idea on how the user would switch between the 3 modes. Currently we just toggle between 1 and 2 with `;;repeat`

knight-ryu12:
I'd say, need to make ;;repeat argumented.

Frederikam:
Yeah, that's what I'm thinking, but what would the usage be like? What do we call modes 2 and 3?

knight-ryu12:
that'd be better.
Just 2,3 toggling is ok

knight-ryu12:
can you tell us how queue is handled?

Frederikam:
I could but my classes have just started again. I'll get back to you when I
have the time.

Frederikam:
## How the audio queue works
The queue in FredBoat is partially held in the `GuildPlayer` and the `SimpleTrackProvider`. When an `AudioTrack` is queued, commonly by the `AudioLoader`, the track is added to the provider's internal `ConcurrentLinkedQueue`. This queue is completely independent of shuffle. The shuffled list is sorted when it is requested and cached until the provider is modified.

The SimpleTrackProvider's purpose is to provide audio tracks upon request by the `GuildPlayer`, which keeps the currently playing (or paused) `AudioTrack` for itself. If the end of a track is reached (eg not stopped/skipped) and the player is in the current repeat mode, the track is simply restarted.

Frederikam:
What I'm thinking is this:
```
;;repeat off // Disables repeat
;;repeat single // Repeats a single track until skipped
;;repeat all // Ended tracks are moved back into the queue
;;repeat // Displays usage information
```

napstr:
Why not both?

I understand the two modes not as exclusive.
`repeat all` boils down to never kicking a song out of the queue, instead putting it back in after it has finished playing.
That's not exclusive to having the first song in front of the queue playing over and over again.

imo solve this by just adding a `;;repeatall` command that will cause finished tracks to get added to the back of the queue.

Let me know if my proposal makes sense or you prefer the three-mode thing, so I can implement the preferred solution.

knight-ryu12:
[...]

PR #93, opened 2017-03-12
Repeat whole playlist added

napstr:
related issue: #42

As specified in the issue this adds another repeat mode that will repeat the whole playlist. The handling of the shuffled list has been adapted and is also more consistent now. Additionally, it is now actually shuffled ;-) Before is was using `Collections.sort()` instead of `Collections.shuffle()`.

A message explaining the `repeat` command will be displayed if the command is used in the wrong way.

Language repo has been updated too.

### DIFF
--- a/FredBoat/src/main/java/fredboat/audio/queue/SimpleTrackProvider.java
+++ b/FredBoat/src/main/java/fredboat/audio/queue/SimpleTrackProvider.java
@@ -143,6 +143,11 @@ public class SimpleTrackProvider extends AbstractTrackProvider {
         if (shuffle) shouldUpdateShuffledQueue = true;
     }
 
+    public synchronized void reshuffle() {
+        queue.forEach(AudioTrackContext::randomize);
+        shouldUpdateShuffledQueue = true;
+    }
+
     @Override
     public synchronized List<AudioTrackContext> getAsListOrdered() {
         if (!isShuffle()) {
--- a/FredBoat/src/main/java/fredboat/audio/queue/SimpleTrackProvider.java
+++ b/FredBoat/src/main/java/fredboat/audio/queue/SimpleTrackProvider.java
@@ -25,7 +25,9 @@
 
 package fredboat.audio.queue;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import java.util.concurrent.ConcurrentLinkedQueue;
 
 public class SimpleTrackProvider extends AbstractTrackProvider {
@@ -37,7 +39,7 @@ public class SimpleTrackProvider extends AbstractTrackProvider {
 
     @Override
     public AudioTrackContext getNext() {
-        if(!isShuffle()) {
+        if (!isShuffle()) {
             return queue.peek();
         } else {
             return getAsListOrdered().get(0);
@@ -46,19 +48,27 @@ public class SimpleTrackProvider extends AbstractTrackProvider {
 
     @Override
     public AudioTrackContext provideAudioTrack(boolean skipped) {
-        if (isRepeat() && !skipped && lastTrack != null) {
+        if (getRepeatMode() == RepeatMode.SINGLE && !skipped && lastTrack != null) {
             return lastTrack.makeClone();
         }
+        if (getRepeatMode() == RepeatMode.ALL && lastTrack != null && !isShuffle()) {
+            //add a fresh copy of the last track back to the queue, if the queue is being repeated
+            queue.add(lastTrack.makeClone());
+        }
         if (isShuffle()) {
-            //Get random int from queue, remove it and then return it
-            List<AudioTrackContext> list = getAsListOrdered();
+            List<AudioTrackContext> updatedShuffledQueue = getAsListOrdered();
 
-            if (list.isEmpty()) {
+            if (updatedShuffledQueue.isEmpty()) {
                 return null;
             }
 
-            shouldUpdateShuffledQueue = true;
-            lastTrack = list.get(0);
+            if (getRepeatMode() == RepeatMode.ALL && lastTrack != null) {
+                AudioTrackContext clone = lastTrack.makeClone();
+                updatedShuffledQueue.add(clone);
+                queue.add(clone);
+            }
+
+            lastTrack = updatedShuffledQueue.remove(0);
             queue.remove(lastTrack);
             return lastTrack;
         } else {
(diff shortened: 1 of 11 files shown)


FredBoat/src/main/java/fredboat/audio/queue/SimpleTrackProvider.java
```diff
@@ -126,25 +135,29 @@ else if (remain <= 0) {
...
         List<AudioTrackContext> newList = new ArrayList<>();
 
         //Update the new queue
-        int i = 1;
         for (AudioTrackContext atc : getAsList()) {
             newList.add(atc);
-            i++;
         }
 
-        Collections.sort(newList);
```
Frederikam: We are *intentionally* using `#sort()`. By replacing this with `#suffle()` you're causing the list to reshuffle every time the queue is accessed after a modification to the queue.

napstr: Yeah, but then repeat all and shuffle won't be possible to run together (which would be a shame), because the player will just be switching between the two topmost tracks that `sort()` puts there. It's kinda a buggy workaround. I can implement a true shuffled queue that doesn't get reshuffled on every operation.
For example, in this PR, already the queue is **not** reshuffled after a song has finished playing.
It does get shuffled on all those other operations though, like adding and removing songs (and maybe other stuff, need to check the code of the `SimpleTrackProvider`). If this is not the expected behaviour I can change that appropriately, just let me know what behavior would be expected for adding (add tracks to the end of the shuffled queue or shuffle them in?) and removing tracks (reshuffle the queue or not?).

Frederikam: If you want to reshuffle a track into the list, just change it's `rand` value. This will effectively give the track a random index.

napstr: Oh I see. You never saw me call the shuffle thing broken. homerbackingintobushes.gif
I'll let you know when I have adjusted this.

napstr: For simplicity sake I'm assuming rand values are 0-100.

regular playlist:

Nr Name     Rand 

01 TrackA  20
02 TrackB  35
03 TrackC  60
04 TrackD  5
05 TrackE  94
06 TrackF  42
07 TrackG  78
08 TrackH  17



after shuffling:

Nr Name     Rand

01 TrackD  5
02 TrackH  17
03 TrackA  20
04 TrackB  35
05 TrackF  42
06 TrackC  60
07 TrackG  78
08 TrackE  94


After Track D has played, there is no sane way to add it back to the shuffled queue.
If we just request a new rand for every played track and add it to the queue at that position:

Nr Name     Rand
 
01 TrackH  17
02 TrackA  20
03 TrackB  35
04 TrackF  42
**05 TrackD  51**
06 TrackC  60
07 TrackG  78
08 TrackE  94


So for Track E to be played during `repeat all` and `shuffle` _every other_ track has to roll a rand >94 which is extremely unlikely.


I don't see a sane way to solve this with the current implementation of the shuffled list.


The current implementation has weaknesses even without using repeat. Imagine this situation after playing half of the tracks from the list above:

Nr Name     Rand
 
01 TrackD  51
02 TrackC  60
03 TrackG  78
04 TrackE  94

Adding a new track to this playlist has a probability of ~50% for it being randed into the first position of the shuffled list, because tracks with rands of 0-50 will land in front of the other tracks who have randed higher values. In my opinion tracks added to a shuffled playlist should have a random chance of showing up anywhere on it.


Proposal: rework the shuffle system of `SimpleTrackProvider`

introduce `shuffledQueue` as a concurrent queue that will also allow entries to be added in the middle (synchronized array list ?)

update `shuffledQueue` with every update of `queue`:
- add tracks to a random position in `shuffledQueue`
- remove tracks in the same way they are removed from `queue`

have these operations done in a synchronized way for thread safety

the `shuffle` command will then just switch between these two queues. recalculations aren't necessary because both lists will be kept updated parallelly


additional command idea: `reshuffle` to force a shuffle of the `shuffledQueue`



Let me know what you think.

Frederikam: Yeah, I see the problem here. I have been suspicious why this was happening, but I'm surprised I didn't realise the cause earlier. I don't really understand what you are proposing, but here is my idea:

When adding a new track to a shuffled queue, iterate this on each track in the shuffled queue:

image


The index `i` would be the index the track would have after being shuffled. This would result in an even distribution without altering how the user sees the shuffled track. See this example with track 0 and 8 in a queue with 9 tracks:
image
When the next track is added, each interval has the same size, thus we ensure unbiased random placement.

napstr: That solves it.
To handle `repeat all` + `shuffle` we'll just set the (clone of the) just played tracks' rand value to `1.0`, which will result in the track showing up at the back of the shuffled queue.
For that to be possible we slightly alter your proposed formula to `queueSize + 2`, otherwise the last track would always get a 1.0, which should be a value reserved for adding a repeated track to the back of the shuffled queue.

With the formula guaranteeing that the values stay between 0 < rand < 1.0 we multiply them with `Integer.MAX_VALUE` to keep the rand value an `int`.

Frederikam: But that isn't the expected behaviour. I would expect an ended track to be randomly shuffled into the queue. To achieve that, it would be as simple as resetting the `rand` value.

my proposed formula would never result in 1. In my example, I showed the calculation of the first and last track. Notice how the last track resulted in `0.9`.

I have no idea why you want to ensure that the rand is an int.

napstr: If you shuffle the just played track back into the queue you have the same problem of tracks who have been initially randed to the back of the queue almost never actually being played. This is more problematic the bigger the playlist is.
You are correct about the formula.
Some other class was getting the rand int (don't know what for right now) so I didn't want to change that.

Frederikam:
The problem with the remaining class not getting played because of a high
rand value would be solved by using that formula.

`TrackEndMarkerHandler` is also using the rand value, but that is only to
check for equality. I realise I should have used `AudioTrackContext#equals()`
instead.

Frederikam:
Let me know when you want me to review this

napstr:
Been testing stuff, everything looks fine now. You can go ahead.

Behavior with `shuffle` on and `repeat all`: New tracks will be shuffled anywhere into the list, tracks that just played will be put at the end of the shuffled queue.
Users can reshuffle the queue with the reshuffle command.

Watch out for the double skip bug.

Frederikam:
This works really great! There's just one more thing for the sake of quality. When a user invokes ;;reshuffle and the queue isn't in shuffle mode the bot should refuse and warn the user. You'll need another I13n PR.

napstr:
Done.

PR #111, opened 2017-03-18
Require the User to be in a voice channel to use Play at all

### LINKED ISSUE #104: Say “You are not in a voice channel” earlier
"Just a little something that annoys me that would be great to fix. If a player is not in a voice channel, and they do ;;play [search], after they choose THEN it says “You are not in a voice channel”. Why not say that before doing the YT search? It saves time, and reduces the YT search count."

Picked this feature request up from the #suggestions channel

Nanabell:
image
1. not in voice
2. in voice

This resolves #104


Nanabell:
Also you should update the self-hosting guide to mention that you will need to include at least the en_US in the resources\lang folder and you also should change the scope to run it since its set to 0x111 which includes the self-bot part.

Frederikam:
Nah, you just need to clone with `--recursive` to get the files. Also the selfbot thing is deprecated and will be removed.

PR #116, opened 2017-03-23
Update to new prefix

gtmx1234:
Changed to standard perfix in config file

### DIFF
--- a/index.md
+++ b/index.md
@@ -13,51 +13,51 @@ __If you want music__ you should be using https://discordapp.com/oauth2/authoriz
 
 | Command                    | Description                                                           | Example usage                               |
 |----------------------------|-----------------------------------------------------------------------|---------------------------------------------|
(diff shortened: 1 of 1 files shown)


Frederikam:
The primary audience of those docs is for users who are using the official hosted version of FredBoat, which is used on over 85k servers, where the prefix is `;;`.

napstr:
Hey @gtmx1234 , thanks for taking the time and working on something that you consider important. 
Your contribution didn't target the configuration file, but the documentation of FredBoat. Obviously the documentation should cater to FredBoats target audience, which Fred described above. 
Nevertheless we are working on allowing custom prefixes an every Discord guild which I assume will solve your problem - stay tuned for updates. 
We have several other issues open for contribution, so if you want to help, feel free to work on anything from here: https://github.com/Frederikam/FredBoat/issues
Happy coding!

Issue #108, opened 2017-03-18
pastebin support for playlists


napstr:
"make fredboat also import playlists from pastebin.com posts like http://pastebin.com/fDyRnGXg"

`PlaylistImportSourceManager` does it for Hastebin playlists. Good starting point when writing code to do it for pastebin.

Picked this feature request up from the #suggestions channel

pmpeev:
I am working on it.

PR #118, opened 2017-03-24
Added support for loading playlists from Pastebin

pmpeev:
I implemented simple option to load files from pastebin.com as for #108. If other paste services are needed in the future, they could be easily added by adding a regex constant for the URL of the desired service and modifying the corresponding switch statement.


Frederikam:
Works like a charm. Thanks! 👍

Issue #73, opened 2017-02-09
Music can not be bent.


suiginn:
I'm a beginner program.
"Http://docs.frederikam.com/selfhosting" I did it as written on the site, but music did not flow
How do I get music to flow?

suiginn:
Currently we have confirmed the following errors.

Exception in thread "JDA AudioConnection Guild: 268663931692777486 Ready Thread" java.lang.UnsatisfiedLinkError: Required library at /natives/darwin/libudpqueue.dylib was not found
At com.sedmelluq.discord.lavaplayer.natives.NativeLibLoader.extractLibrary (NativeLibLoader.java: 93)
At com.sedmelluq.discord.lavaplayer.natives.NativeLibLoader.load (NativeLibLoader.java: 77)
At com.sedmelluq.discord.lavaplayer.udpqueue.natives.UdpQueueManagerLibrary.getInstance (UdpQueueManagerLibrary.java: 13)
At com.sedmelluq.discord.lavaplayer.udpqueue.natives.UdpQueueManager. <Init> (UdpQueueManager.java: 26)
At com.sedmelluq.discord.lavaplayer.jdaudp.NativeAudioSendFactory.initialiseQueueManager (NativeAudioSendFactory.java: 31)
At com.sedmelluq.discord.lavaplayer.jdaudp.NativeAudioSendFactory.addInstance (NativeAudioSendFactory.java: 58)
At com.sedmelluq.discord.lavaplayer.jdaudp.NativeAudioSendSystem.start (NativeAudioSendSystem.java: 22)
At net.dv8tion.jda.core.audio.AudioConnection.setupSendSystem (AudioConnection.java: 225)
At net.dv8tion.jda.core.audio.AudioConnection.access $ 200 (AudioConnection.java: 47)
At net.dv8tion.jda.core.audio.AudioConnection $ 1.run (AudioConnection.java: 123)

Frederikam:
It looks like you are trying to host FredBoat on a Mac. FredBoat depends on https://github.com/sedmelluq/lavaplayer to play music, which does not seem to support Darwin (the OSX kernel).

This normally wouldn't be a problem for a Java library, but this library isn't pure Java.

sedmelluq:
LavaPlayer itself works on OS X, but the native audio packet send system doesn't. Remove `.setAudioSendFactory(new NativeAudioSendFactory())` part in `FredBoatBot` class if you want to run it on OS X.

Frederikam:
[...]

PR #123, opened 2017-04-01
Fixed audio system for Mac self-hosters.

kitkatdesu:
Related to #73. Tested on macOS Sierra 10.12.2.


Frederikam:
I don't own a mac so thanks for finding and fixing this 👍

kitkatdesu:
IDEA looks great on Mac. 
125am

PR #128, opened 2017-04-04
Allow users and roles configurated as admins to invoke IAdminCommands

napstr:


FredBoat/src/main/java/fredboat/commandmeta/CommandManager.java
```diff
@@ -86,7 +80,26 @@ public static void prefixCalled(Command invoked, Guild guild, TextChannel channe
             return;
         }
 
-        if (invoked instanceof ICommandOwnerRestricted) {
+        if (invoked instanceof IAdminCommand) {
```
Frederikam: Is it just me or are commands with `ICommandOwnerRestricted` AND `IAdminCommand` effectively restricted to admins? Shouldn't these two `if else` code blocks be switched around?

napstr: Yes and no. The code is correct but it looks horrible.

Logic of this block is like this:
If the command is an IAdminCommand, stop everyone who is neither the bot owner nor an admin
else if it's a ICommandOwnerRestricted, stop everyone who is not the bot owner

This can be written in a nicer way if we assume IAdminCommands are always also ICommandOwnerRestricted which is a) currently always true (no need to allow commands for admins if they aren't owner restricted in the first place) and b) beyond yesterday me's understanding.

brb

napstr:
@Frederikam should look less weird now.

FredBoat/src/main/java/fredboat/commandmeta/CommandManager.java
```diff
@@ -88,9 +82,12 @@ public static void prefixCalled(Command invoked, Guild guild, TextChannel channe
 
         if (invoked instanceof ICommandOwnerRestricted) {
```
Frederikam: Wait, does this assume that all commands that are `IAdminCommand` can be run by anyone if they are not `ICommandOwnerRestricted`?

In my opinion, I believe the two interfaces should be mutually exclusive.

`IAdminCommand` can only be used by admins (and the owner is implicitly an admin).
`ICommandOwnerRestricted` can only be used by the owner.

napstr: Sure, mutually exclusive is possible, and less error prone for the future. There were raisins I didn't like that at first, but nvm it's fine.

brb

napstr:
Third time's the charm?

Frederikam:
Much better! I'm assuming you've already tested this. Really don't want to expose these commands.

PR #134, opened 2017-04-09
Database fix yo

napstr:


Frederikam:
What was the problem though?

napstr:
Never calling .close() on the EntityManagers and keeping them around for every thread.
Don't need to do that if you're using a pool like Hikari.

Frederikam:
That would have been nice to know a few months ago. Thanks for the help!

napstr:
It's not wrong to do it the way it was done, actually it's on the right path things need to be when you don't have a pool. This pool thing was new to me too, and not exactly greatly document _what_ it is that you have to do to use it in our specific scenario. I've also read tons of stuff up, and found a setting that will create and update the tables we use automatically, which I find incredibly handy. Check this: 607d6d6
There is some kind of danger label attached to it though, rule of thumb: don't change the tables between redeploys too crazily, just extend them (probably stuff like don't add NOT NULL columns to existing tables, that might be weird).

PR #143, opened 2017-04-13
Look up a members voice chat in a less awkward way

napstr:


napstr:
I suspected as much.

Any chances the bug is still there?

Frederikam:
Most likely, as it's likely the cause that the bot can't see users in voice chats.

PR #148, opened 2017-04-16
Fallback to SQLite db if no jdbcUrl is given

napstr:
This is aimed at selfhosters.


napstr:
While core functionality works fine (saving and retrieving GuildConfigs), I just noticed the ;;test command fails with this. I'll investigate that first.

napstr:
Yep.

PR #149, opened 2017-04-17
Validate logins earlier + fix NPE

napstr:
Do it while creating the shards, not afterwards (which can take 10 minutes)


napstr:
I ran into the trap of assuming JDA can get a hold of every Discord user ever, while that's not true. It can only get a hold of users the bot shares servers with. AudioTrackContext is hopefully fixed for real this time.

PR #156, opened 2017-04-22
Added weather command.

HgGamer:


FredBoat/src/main/java/fredboat/Config.java
```diff
@@ -184,6 +188,7 @@ public Config(File credentialsFile, File configFile, int scope) {
 
             imgurClientId = (String) creds.getOrDefault("imgurClientId", "");
 
+	    openweathermapKey = (String) creds.getOrDefault("openweathermapKey", "");
```
Frederikam: Please indent your code properly. Also configure your IDE to use 4 spaces instead of tabs.

FredBoat/pom.xml
```diff
@@ -164,6 +164,11 @@
             <artifactId>jsch</artifactId>
             <version>0.1.54</version>
         </dependency>
+	<dependency>
+	    <groupId>com.googlecode.json-simple</groupId>
+	    <artifactId>json-simple</artifactId>
+	    <version>1.1.1</version>
+	</dependency>
```
Frederikam: We already have `json.org`. Would be great if you could use that instead of adding a new dependency.

FredBoat/src/main/java/fredboat/command/util/WeatherCommand.java
```diff
@@ -0,0 +1,103 @@
...
+	if (args.length < 2) {
+	    String command = args[0].substring(Config.CONFIG.getPrefix().length());
+	    HelpCommand.sendFormattedCommandHelp(guild, channel, invoker, command);
+	    return;
+        }
+        String res = message.getRawContent().substring(args[0].length() + 1);
+        Message myMsg;
+        try {
+
+            myMsg = channel.sendMessage('\u200b'+ LoadWeather(res)).complete(true);
```
Frederikam: Please use RestAction#queue() instead. We don't need to wait for the message to return. This also avoids the need for a try/catch

napstr:
Hey, if you are using IntelliJ,an easy way to fit in with the existing code style would be to use the
[Save Actions plugin for IntelliJ](https://plugins.jetbrains.com/plugin/7642-save-actions) and configure it like this:

image

And this is how the indent settings look:
image

HgGamer:
Thanks for your support. Currently I'm stuck with unit testing, but as soon as i have a bit free time i'll fix the code and send a new pull request.

Frederikam:
Don't send a new pull request though. Just push more commits and they will show up here.
image

HgGamer:
Okay, thanks. I'm new to git.

Frederikam:
It's been 60 days now. Do you mind if i close this, or do you want to finish it?

HgGamer:
Feel free to close it, i'm tired of java. Currently working on a nodejs based bot with one of my friends.

Issue #109, opened 2017-03-18
Ratelimits for FredBoat


napstr:
Ratelimit for users. Especially stuff that does a lot of requests/heavy calculation stuff (example: loading a spotify playlist)

This needs some kind of (temporary) data keeping on users. What kind of DB is the current one used?

Nanabell:
postgresql iirc

Frederikam:
I don't see why this could not be achieved from just a HashMap?

napstr:
Some general thoughts:
- Rate limiting on a per-guild level and a (global) per-user level -> take out abusers without affecting the guilds they use for their actions
- Even if a user/guild has depleted credits to add songs, music control commands should keep working.
- I don't really want to stop someone from loading some crazy 500 songs or whatever playlist, but if he decides to do, he should be blocked for an appropriate time from adding more songs.
- There are two possible ways of abuse: spamming expensive commands (like loading songs), and just spamming a lot of commands (no one needs to pause/unpause or request meme pictures 100 times in a minute). The latter one should not block for a long time (especially not for guilds), some per-minute rate of commands seems appropriate.
- song limit for the guilds should replenish more often and be rather generous  than the one for a single user. we want to punish individual abusers in the first place, not the guild they are home to. guilds should only be punished when they are home to several abusers that they aren't taking care of.

Leaves me with a fourfold system of rate limiting:
(all numbers made up, I think you have more experience than me to adjust them properly; it can be done when the code is written, this is more about exploring what kind of rate limiting we want)

song limit per user (adding 100 songs per hour, updated every hour)
song limit per guild (adding 1000 songs per hour, updated every 10 mins)
command limit per user (issuing 100 commands per minute, updated every minute)
command limit per guild (issuing 200 commands per minute, updated every minute)


Some examples:
- A user adds a 500 song troll playlist (we should still add some kind of hard limit to playlists, maybe only allow moderators on a guild add really long ones): he will have to wait 5 hours until he can add more songs, while his guild still has enough limit left to use the bot
- A spammer spams ;;np or whatever commands and gets limited at 100 commands per minute; other users of the guild still have access to 100 commands per minute to use; if it's a guild of spammers they get limited at 200 commands per minute, and all users get affected, even if it's only until the minute ends to give guild admins just enough motivation to get rid of the abusers without making it unusable for everyone else entirely.

Additional things:
[...]

PR #165, opened 2017-04-25
Add ratelimiter and autoblacklist

napstr:
Based on #109 

This isn't finished yet, I need input about the UX side of this.


FredBoat/config.yaml
```diff
@@ -3,4 +3,5 @@ patron: false
 development: true        # Set this to true for self hosting
 prefix: '<<'             # Default prefix used by the bot
 restServerEnabled: true  # Change this if you are running multiple FredBoat bots on the same machine
-admins: []               # add comma separated userIds and roleIds that should have access to admin commands
\ No newline at end of file
+admins: []               # add comma separated userIds and roleIds that should have access to admin commands
+useAutoBlacklist: false   # set to true to automatically blacklist users who frequently hit the rate limits
```
Frederikam: Why not enable this by default? Too agressive?

FredBoat/src/main/java/fredboat/util/ratelimit/SlowImportedPlaylistRateIdentifier.java
```diff
@@ -0,0 +1,9 @@
+package fredboat.util.ratelimit;
+
+/**
+ * Created by napster on 24.04.17.
+ * <p>
+ * This thing is a helper class for the Ratelimiter
+ */
+public class SlowImportedPlaylistRateIdentifier {
```
Frederikam: What is the purpose of this class?

FredBoat/config.yaml
napstr: No reason really. I don't find it aggressive at all. If we properly communicate why it is happening (UX thing) I would turn it on by default.

Frederikam: Ah sorry, I didn't read that this was unfinished. When we do that it would be a good idea to enable the blacklist by default.

FredBoat/src/main/java/fredboat/util/ratelimit/SlowImportedPlaylistRateIdentifier.java
napstr: This is one of the things I'm looking into optimizing away.

Currently the Ratelimiter works with objects to identify the scope of the requests and apply rate limits, which is fine and intuitive for commands, but not for other stuff.

napstr:
> Can you summarise how this behaves?

Sorry, missed this yesterday.

Ratelimits as proposed here: https://github.com/Frederikam/FredBoat/issues/109#issuecomment-292219229 except for the playlist/user limit which we deemed suboptimal.

When a user hits a ratelimit he will be mentioned and told that he is (currently with no additional info, just something along the lines "you hit a a rate limit")

If the user hits it X times (for example 5) (TODO: make sure this is only triggered if hits happen in a short time period), he will be blacklisted. First time that happens for 1 minute, then 10 minutes, then 1 hour, 1 day, 1 week. It caps out at 1 week blacklisting.
I didn't bother translating all of these into config options, they are hardcoded, but they can be freely adjusted so if you think that's too harsh/not enough let me know and we can change that. You can find the Blacklist levels at the top of the Blacklist class.

When a user gets blacklisted, he will be mentioned and told the length of the blacklist.

napstr:
Optimization and DB persistence done, only UX/i18n left.

napstr:
This thing is ready to be pulled (ignore the failed build, I forgot to update the i13n project first).

Just one question left: On which distributions should this be enabled? I kinda want to disable it for Patron and Development.

Frederikam:
I don't see why you would distance development from music even more than it is. Also I think we should enable it for patron as well. I know, these people paid for the bot, but that doesn't mean they can ruin the fun for others.

napstr:
Ok then. I'm thinking about the development distribution as the self hoster full one. Some people are self hosting the full bot. I agree with you that abusing the development distribution is bad. 

Then this is ready. Please merge or let me know if it's OK to push this to development.

knight-ryu12:
Um.
at last commit, Travis saying No.
Maybe fix it?

napstr:
Triggering Travis rebuild.

napstr:
Done.

PR #190, opened 2017-06-16
Compact command registry

Shredder121:
I noticed while hardcoding in more aliases, that typos can be made a little too easily.
By including it in the call to `registerCommand` you don't have to type the command twice, `3`x, `4`x, `n`x.

Also, let me know if the original parameter ordering should be preserved, or the one in my first commit is better, should you want to merge this.
(then I can either drop or squash the last commit)


Shredder121:
Signed 👍

napstr:
Looks good to me.

Frederikam:
This looks great! I like how you rearranged the parameters. It makes things more readable.

Ritchie454:
This makes my slight OCD very happy, nice job 👍

Shredder121:
> This looks great! I like how you rearranged the parameters.
i could have cleaned up the commits, so it was actually rearranged, or did you mean that it was fine like that?

Shredder121:
*I secretly have a few other outstanding changes, but didn't get to commit them for some time.*
*But good to hear it is being received so well!*

Frederikam:
>I could have cleaned up the commits, so it was actually rearranged, or did you mean that it was fine like that?

GitHub gives me the ability to squash before merging, which I did.

Shredder121:
Then I assume the ordering is fine as it is now, good good. 👍

PR #191, opened 2017-06-16
Use an executor for the processing of the agents

Shredder121:
I wanted to wait with this one last, but I think it's also good to get some comments/feedback in.
So here goes.

This makes the point where agents are started unified, and cleans up the `//noinspection InfiniteLoopStatement while (true) {`, each one start a Thread, (correctly) configuring it, etc. mantra.

Caveats/notes:

- [x] This makes the starting/handling/rate-scheduling/stopping universal.
- [x] `log.info("starting xxx")` should be either 1) removed, or 2) be accompanied by `log.info("finished xxx")` or 3) be changed in logging level.
- [x] Decide on whether to use `scheduleWithFixedDelay` (which was more akin to the previous code, since there were always fixed sleep intervals) or `scheduleAtFixedRate` (which makes more sense (?) when dealing with agents/cleanup units of work).
- [x] Deal with potential `Throwable`s killing off the agents.
- [ ] Decide whether a single worker thread is enough, or if a larger threadpool or a cached threadpool is better, so 2, 3, 4 threads? Or let the cached threadpool handle it?
- [ ] Decide whether we use `AGENTS.awaitTermination()` in the shutdown, or do we just let it shutdown while we do other shutdown-y things?
- [ ] Needs a lot of eyes on the code. I have been running this code for about/over a week or two, but that's different from when it would go live for thousands of servers.
- [ ] Do not merge until all checks have been ticked, since other issues can be raised 😨. (I don't want to be *that guy who broke everything*)

### DIFF
--- a/FredBoat/src/main/java/fredboat/agent/FredBoatAgent.java
+++ b/FredBoat/src/main/java/fredboat/agent/FredBoatAgent.java
@@ -37,7 +37,7 @@ public abstract class FredBoatAgent implements Runnable {
     private static final String IDLE_NAME = "idle agent worker thread";
     private static final String RUNNING_NAME = "%s agent worker thread";
 
-    private static final ScheduledExecutorService AGENTS = Executors.newSingleThreadScheduledExecutor(runnable -> {
+    private static final ScheduledExecutorService AGENTS = Executors.newScheduledThreadPool(2, runnable -> {
         Thread thread = new Thread(runnable, IDLE_NAME);
         thread.setPriority(4);
         return thread;
--- a/FredBoat/src/main/java/fredboat/agent/FredBoatAgent.java
+++ b/FredBoat/src/main/java/fredboat/agent/FredBoatAgent.java
@@ -68,9 +68,9 @@ public abstract class FredBoatAgent implements Runnable {
         LAST_RUN_TIME.put(this.getClass(), System.currentTimeMillis());
         try {
             Thread.currentThread().setName(name);
-            log.info("running");
+            log.debug("running");
             doRun();
-            log.info("done");
+            log.debug("done");
         } catch (Throwable t) {
             log.warn("Whoa! Unhandled throwable!", t);
         } finally {
--- a/FredBoat/src/main/java/fredboat/agent/FredBoatAgent.java
+++ b/FredBoat/src/main/java/fredboat/agent/FredBoatAgent.java
@@ -28,6 +28,9 @@ package fredboat.agent;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
@@ -36,6 +39,8 @@ public abstract class FredBoatAgent implements Runnable {
 
     private static final String IDLE_NAME = "idle agent worker thread";
     private static final String RUNNING_NAME = "%s agent worker thread";
+
+    private static final Map<Class<? extends FredBoatAgent>, Long> LAST_RUN_TIME = new ConcurrentHashMap<>();
 
     private static final ScheduledExecutorService AGENTS = Executors.newScheduledThreadPool(2, runnable -> {
         Thread thread = new Thread(runnable, IDLE_NAME);
@@ -60,6 +65,7 @@ public abstract class FredBoatAgent implements Runnable {
 
     @Override
     public final void run() {
+        LAST_RUN_TIME.put(this.getClass(), System.currentTimeMillis());
         try {
             Thread.currentThread().setName(name);
             log.info("running");
@@ -76,6 +82,10 @@ public abstract class FredBoatAgent implements Runnable {
 
     public static void start(FredBoatAgent agent) {
         AGENTS.scheduleAtFixedRate(agent, agent.millisToSleep, agent.millisToSleep, TimeUnit.MILLISECONDS);
+    }
+
+    public static Map<Class<? extends FredBoatAgent>, Long> getLastRunTimes() {
+        return Collections.unmodifiableMap(LAST_RUN_TIME);
     }
 
     public static void shutdown() {
--- a/FredBoat/src/main/java/fredboat/agent/FredBoatAgent.java
+++ b/FredBoat/src/main/java/fredboat/agent/FredBoatAgent.java
@@ -82,6 +82,7 @@ public abstract class FredBoatAgent implements Runnable {
 
     public static void start(FredBoatAgent agent) {
         AGENTS.scheduleAtFixedRate(agent, agent.millisToSleep, agent.millisToSleep, TimeUnit.MILLISECONDS);
+        LAST_RUN_TIME.put(agent.getClass(), 0L);
     }
 
     public static Map<Class<? extends FredBoatAgent>, Long> getLastRunTimes() {
--- a/FredBoat/src/main/java/fredboat/agent/FredBoatAgent.java
+++ b/FredBoat/src/main/java/fredboat/agent/FredBoatAgent.java
@@ -1,0 +1,82 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2017 Frederik Ar. Mikkelsen
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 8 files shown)


FredBoat/src/main/java/fredboat/agent/FredBoatAgent.java
```diff
@@ -32,7 +32,7 @@
 public abstract class FredBoatAgent implements Runnable {
 
     private static final ScheduledExecutorService AGENTS = Executors.newSingleThreadScheduledExecutor(runnable -> {
-        Thread thread = new Thread(runnable);
+        Thread thread = new Thread(runnable, "agent runner thread");
```
Shredder121: Stub, if anyone knows of a *nicer* name?

Frederikam:
I'd say `scheduleAtFixedRate` is fine as our agents all take a long while between executions. The time they run is still much smaller than in the production environment.

FredBoat/src/main/java/fredboat/agent/FredBoatAgent.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+public abstract class FredBoatAgent implements Runnable {
+
+    private static final ScheduledExecutorService AGENTS = Executors.newSingleThreadScheduledExecutor(runnable -> {
+        Thread thread = new Thread(runnable, "agent runner thread");
+        thread.setDaemon(true);
```
Shredder121: Also, one could argue since we now properly shutdown the agents, that we do not need the worker(s) to be daemon threads.

Frederikam: 👍

FredBoat/src/main/java/fredboat/agent/FredBoatAgent.java
```diff
@@ -0,0 +1,57 @@
...
+package fredboat.agent;
+
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+public abstract class FredBoatAgent implements Runnable {
+
+    private static final ScheduledExecutorService AGENTS = Executors.newSingleThreadScheduledExecutor(runnable -> {
+        Thread thread = new Thread(runnable, "agent runner thread");
```
Shredder121: In fact **not** outdated, *GitHub*, still a stub, if someone knows a better name.

Or a naming scheme, if we go for multiple threads in the thread pool.

napstr:
So this uses one Thread to do everything? We've had the situation that a thread got stuck doing his task. What happens here in that situation?

Frederikam:
>So this uses one Thread to do everything? We've had the situation that a thread got stuck doing his task. What happens here in that situation?

We'll probably want to use a pool of threads. The production bot already has several thousands of threads running. A few more won't hurt.

napstr:
Actually in that situation we had to go over a thread dump and find the threads by name. So I think it's important to keep the agent's tasks independent from each other, and have a distinct name.

Frederikam:
Can we not just change the name when a task starts? We can just add it to the constructor of the abstract agent.

Shredder121:
>A few more won't hurt.

That.. is a way of thinking about it.

> We've had the situation that a thread got stuck doing his task. 

> So I think it's important to keep the agent's tasks independent from each other, and have a distinct name.

Yes distinct threads is a good idea.

> Can we not just change the name when a task starts? We can just add it to the constructor of the abstract agent.

Yeah, we can do that, I'm now changing the code, and testing it again while listening to the music it produces.

Shredder121:
> Actually in that situation we had to go over a thread dump and find the threads by name. So I think it's important to keep the agent's tasks independent from each other, and have a distinct name.

We can also introduce a ThreadGroup, so the spawned threads are grouped?
But maybe later, now just correctness.

Shredder121:
Thread group is maybe not necessary, but at least the name now reflects the agent that's running.

Frederikam:
ea3e280082429d36644e9baf146372ce19c7607b causes a conflict with this PR

Shredder121:
I'll rebase later, now need to do something else.

But thanks for letting me know 😄

napstr:
This is important:

http://code.nomad-labs.com/2011/12/09/mother-fk-the-scheduledexecutorservice/

TL;DR: If scheduled tasks encounter an exception, they will be cancelled _silently_.

Shredder121:
I know, that's why I already kept the try-catches.
The only place I didn't is the CarbonitexAgent, because the only play it could throw was already try-catched.
Or do you want to have it anyway?

napstr:
Idk. Things failing silently by default really rustles my jimmies
```
try {}
catch (Throwable t) {}
```
¯\\\_(ツ)\_/¯

Shredder121:
`Throwable`? You mad man!
(But I can do that :/)

knight-ryu12:
ffs use Exception e, It is good for you

Frederikam:
Why silence the exceptions? It's a pain to debug if we don't have a trace.

Shredder121:
How is this?

Shredder121:
The individual agents already have try-catches, so this is just a catch-all for if something goes _horribly_ wrong.

napstr:
Hey idk if this is overkill, but a command to check the status of all FredBoat agents, and to cancel and start them, wouldn't that be neat to have for us admins?

napstr:
Not necessarily a command, I would take colored indicators on some kind of admin dashboard if I could, but that's not a thing.

Shredder121:
Ohh yeah I can whip something up

napstr:
About those uncaught exceptions, TIL there is this thing:

https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.UncaughtExceptionHandler.html

Shredder121:
Yeah, that's useful if they run in a `Thread`, but they run on a thread pool, and I'm pretty sure the worker threads gobble the exception, and don't report it up.
It would be pretty expensive to let the threads actually die after each exception and then create a new one.
<details>
<summary>Snippet taken from `FutureTask.runAndReset()`</summary>

```java
try {
    Callable<V> c = callable;
    if (c != null && s == NEW) {
        try {
            c.call(); // don't set result
            ran = true;
        } catch (Throwable ex) {
            setException(ex);
        }
    }
} finally {
```

</details>

Shredder121:
So, status:
Currently set to use 2 threads, should not have any runaway tasks in the agents(the 'problem size' is pretty small -- playing players, active (potentially dead) shards)

The tasks won't ever die, because we catch all throwables (and at least log it, so we don't shoot ourselves in the foot by silently ignoring any problem)

> Decide whether we use AGENTS.awaitTermination() in the shutdown, or do we just let it shutdown while we do other shutdown-y things?

Do we just assume the pool will have been closed? do we block and wait or not?

P.S.
Did we kill the shard watchdog agent yet? should I pull that?

napstr:
About the shard watchdog: Let's circumvent having to make an ultimate decision and make it a feature flag :upside_down_face: 

We can close the pool in the shutdown hook. Even if it's only running daemons etc that should still be a best practice.

Shredder121:
> We can close the pool in the shutdown hook. Even if it's only running daemons etc that should still be a best practice.

Well it's already closing in the shutdown hook, but do we block and wait until it's finished, or just assume it will go away and do other shutdown-y things in the meantime?
It's doing the latter now, but I want opinions.

Anyway, amount of threads is fine? (1 for the frequent things, the other for backup and having some parallelism?)

napstr:
Put it on a timeout of 30sec or something imo, just in case it gets stuck.
Amount of threads looks fine to me.

Shredder121:
So I rebased this one, but noticed there is not a way anymore to reach the ShardWatchdogAgent.

I see we put it under a feature flag, is that guaranteed safe then?
I assume so, but would like a sign off that it's now good, since we don't have a way to reach it.

napstr:
Looks good to me. I'm just not sure I like the log spam. Probably better to move these to the debug level?
image

Shredder121:
Debug level it is then, one moment.

napstr:
Tbh even the debug level is annoying (which I'm using locally quite often).

Is trace level viable? Or any other idea?

Shredder121:
Trace is doable, sure
Does it end up in any log file?
If not just remove it

napstr:
Nope, we don't log traces at all. This doesn't feel like a useful log anyways since we track the last execution time and it's walled up against exceptions properly too.

Shredder121:
Then I vote for removal of starting/running, but keep the thread name adjustment for if the exceptions happen and something else is logged

PR #200, opened 2017-06-24
Add placefillers in exact format they need to be

Ritchie454:
Seen a few people confused with the formatting so thought adding these changes so stuff is in the exact format may be helpful


Frederikam:
Shouldn't non-required options be left as null?

Nanabell:
i would rather have them shown but add a comment on those which are required saying that it is required

napstr:
Comment the non-required ones out imo.

PR #211, opened 2017-07-02
Change the implementation of most counting commands

Shredder121:
This uses the underlying map, very carefully, which in most cases are safe enough to call.

Note that the '`pom.xml`' that gradle produces has all its dependencies set to `runtime`.
So that means that trove isn't actually a transitive (compile time) dependency of JDA.
That sucks big time, but /shrug.

### DIFF
--- a/FredBoat/feature_flags.properties
+++ b/FredBoat/feature_flags.properties
@@ -31,4 +31,5 @@
 # example to turn a feature on (uncomment it):
 #RATE_LIMITER=false
 #CHATBOT=false
+#DATA_METHODS=false
 PERMISSIONS=false
--- a/FredBoat/src/main/java/fredboat/command/maintenance/ShardsCommand.java
+++ b/FredBoat/src/main/java/fredboat/command/maintenance/ShardsCommand.java
@@ -33,6 +33,7 @@ import it.unimi.dsi.fastutil.longs.LongOpenHashSet;
 import net.dv8tion.jda.core.JDA;
 import net.dv8tion.jda.core.MessageBuilder;
 import net.dv8tion.jda.core.entities.*;
+import net.dv8tion.jda.core.entities.impl.JDAImpl;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -60,8 +61,9 @@ public class ShardsCommand extends Command implements IMaintenanceCommand {
         LongOpenHashSet healthyUsers = new LongOpenHashSet(FredBoat.getExpectedUserCount());
         for (FredBoat fb : shards) {
             if (fb.getJda().getStatus() == JDA.Status.CONNECTED && !full) {
-                healthyGuilds += fb.getJda().getGuilds().size();
-                fb.getJda().getUsers().parallelStream().mapToLong(ISnowflake::getIdLong).forEach(healthyUsers::add);
+                healthyGuilds += fb.getGuildCount();
+                // casting to get the underlying map, this is safe because we only need the .size()
+                ((JDAImpl) fb.getJda()).getUserMap().size();
             } else {
                 if (borkenShards % SHARDS_PER_MESSAGE == 0) {
                     mb = new MessageBuilder()
--- a/FredBoat/src/main/java/fredboat/command/maintenance/ShardsCommand.java
+++ b/FredBoat/src/main/java/fredboat/command/maintenance/ShardsCommand.java
@@ -29,7 +29,6 @@ import fredboat.Config;
 import fredboat.FredBoat;
 import fredboat.commandmeta.abs.Command;
 import fredboat.commandmeta.abs.IMaintenanceCommand;
-import it.unimi.dsi.fastutil.longs.LongOpenHashSet;
 import net.dv8tion.jda.core.JDA;
 import net.dv8tion.jda.core.MessageBuilder;
 import net.dv8tion.jda.core.entities.*;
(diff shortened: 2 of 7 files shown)


Shredder121:
But the ShardsCommand has at least a few calls inlined, because it wants to do the counts incrementally.
What we could do is add them to a second list and then `JDAUtil.countAllGuilds` to only pay the hit once, and the same for `JDAUtil.countAllUniqueUsers`.

Frederikam:
We could also choose to just have ;;mshards show non-unique users

Shredder121:
Rebased, since 1) I forgot to add `JDAUtil` and 2) conflicts.

But do tell me, if the style is okay.
Or if you see more calls that could be changed.

Shredder121:
Anyway, `;;shards` still has the parallelStream, but non-unique might also work.

Shredder121:
Now only to have it in the feature configuration right?

napstr:
Let's not user parallel Streams. I think those are always a bad choice.

Frederikam:
>Let's not user parallel Streams. I think those are always a bad choice.

*Said the guy who used them to break the bot*

😛

napstr:
I'm also the guy who tries to do every mistake exactly once.

Shredder121:
inb4 JDA thinks it's fine:
```java
GuildImpl.java
@Override
public List<Emote> getEmotesByName(String name, boolean ignoreCase)
{
    Args.notNull(name, "name");
    return Collections.unmodifiableList(emotes.valueCollection().parallelStream()
            .filter(e ->
                    ignoreCase
                    ? StringUtils.equalsIgnoreCase(e.getName(), name)
                    : StringUtils.equals(e.getName(), name))
            .collect(Collectors.toList()));
}
```

FredBoat/src/main/java/fredboat/command/maintenance/ShardsCommand.java
```diff
@@ -60,8 +61,9 @@ public void onInvoke(Guild guild, TextChannel channel, Member invoker, Message m
         LongOpenHashSet healthyUsers = new LongOpenHashSet(FredBoat.getExpectedUserCount());
         for (FredBoat fb : shards) {
             if (fb.getJda().getStatus() == JDA.Status.CONNECTED && !full) {
-                healthyGuilds += fb.getJda().getGuilds().size();
-                fb.getJda().getUsers().parallelStream().mapToLong(ISnowflake::getIdLong).forEach(healthyUsers::add);
```
napstr: This line is missing a replacement.
Need to add userIds to healthyUsers,

napstr:
I wouldn't say everything JDA does works well at scale, so sadly it's not necessary a good place to get inspirations from.

FredBoat/src/main/java/fredboat/command/maintenance/ShardsCommand.java
Shredder121: I thought we went the non-unique route.
So in this case we can update a regular long/int right?

If we are to dedupe them, let me know I have an idea.

Shredder121: I thought we went the non-unique route.
So in this case we can update a regular long/int right?

If we are to dedupe them, let me know I have an idea.

FredBoat/src/main/java/fredboat/command/maintenance/ShardsCommand.java
```diff
@@ -58,12 +57,12 @@ public void onInvoke(Guild guild, TextChannel channel, Member invoker, Message m
...
         int borkenShards = 0;
         int healthyGuilds = 0;
-        LongOpenHashSet healthyUsers = new LongOpenHashSet(FredBoat.getExpectedUserCount());
+        int healthyUsers = 0;
         for (FredBoat fb : shards) {
             if (fb.getJda().getStatus() == JDA.Status.CONNECTED && !full) {
                 healthyGuilds += fb.getGuildCount();
                 // casting to get the underlying map, this is safe because we only need the .size()
-                ((JDAImpl) fb.getJda()).getUserMap().size();
+                healthyUsers = ((JDAImpl) fb.getJda()).getUserMap().size();
```
napstr: This should be summing up.

FredBoat/feature_flags.properties
```diff
@@ -31,4 +31,5 @@
 # example to turn a feature on (uncomment it):
 #RATE_LIMITER=false
 #CHATBOT=false
+#DATA_METHODS=false
```
napstr: Fyi this file is almost purely decorative - it needs to be pushed and updated by hand to production.
It's good to keep it in sync with the existing flags, just keep in mind changing it in git changes nothing in production.

FredBoat/src/main/java/fredboat/command/maintenance/ShardsCommand.java
Shredder121: `Actually store it correctly for real this time`
Very good catch >.<

FredBoat/feature_flags.properties
Shredder121: Yeah that would be annoying, I assumed you weren't working from inside the git directory.
Still, keeping it in sync is nice.

napstr: Eventually it will move to the database and a GUI with buttons to click or something.

PR #215, opened 2017-07-03
Count only the current shard's users when asked for it

Shredder121:


Shredder121:
@napstr would you please review?

PR #217, opened 2017-07-04
Fix setting of the frameBufferDuration

Shredder121:
Due to an inconsistency, the frameBufferDuration is capped at 200ms if you want to set a different value.
The default is 5000, but we want it set to 1000.

This means that it actually becomes 200, which on a large scale is not much leeway.

(Don't know if you accept much reflection?)


napstr:
Shouldn't this be patched in lavaplayer? Or is this a hack that only benefits FredBoat?

Shredder121:
Yeah I agree, it should be fixed upstream.

Shredder121:
Hm brain fart?
Just thought that `max` should be fine 😓 
right?

Issue #189, opened 2017-06-15
;;skip command should have a cooldown


Paulloz:
Hi there!  
We just had a dumb issue while using FredBoat: two users sent a `;;skip` command almost at the same time so (obviously) two tracks were skipped.  
Maybe there should be some sort of mechanism preventing `;;skip` commands within a few seconds from one another to execute?

Frederikam:
Never encountered this problem before, but I can see how it could be useful. Maybe a new contributor could handle this as it's a low-hanging apple.

drdgvhbh:
im gonna try working on this issue.

PR #222, opened 2017-07-14
Solving Issue #189

drdgvhbh:
I added a cooldown to the skip command. Its defaulted at 1 second. 
You can change the cooldown using the skip command with this syntax:

image

Where t is the time in milliseconds you want the cooldown to be. And "cd" is the keyword to change the cooldown. You can always change the cooldown and skip at the same time even if the skip command is on cooldown.

I need some help updating the localization for the help message to accommodate the new commands though. I'm also not sure if I should display a message when the user tries to skip when its on cooldown.

### DIFF
--- a/FredBoat/src/main/java/fredboat/command/music/control/SkipCommand.java
+++ b/FredBoat/src/main/java/fredboat/command/music/control/SkipCommand.java
@@ -42,16 +42,48 @@ import net.dv8tion.jda.core.entities.Message;
 import net.dv8tion.jda.core.entities.TextChannel;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
+import org.apache.commons.logging.impl.SLF4JLog;
+import org.slf4j.LoggerFactory;
 
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 public class SkipCommand extends Command implements IMusicCommand, ICommandRestricted {
     private static final String TRACK_RANGE_REGEX = "^(0?\\d+)-(0?\\d+)$";
     private static final Pattern trackRangePattern = Pattern.compile(TRACK_RANGE_REGEX);
+
+    /**
+     * The keyword to change the {@link #skipCooldown}
+     */
+    private static final String COOLDOWN_CMD = "cd";
+
+    /**
+     * The default cooldown for calling the {@link #onInvoke} method in milliseconds.
+     */
+    private static final int DEFAULT_SKIP_COOLDOWN = 1000;
+
+    /**
+     * The latest time {@link #onInvoke} was called in milliseconds.
+     */
+    private long lastInvokeCall;
+
+    /**
+     * The cooldown for calling the {@link #onInvoke} method in milliseconds.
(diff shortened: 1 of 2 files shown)


FredBoat/src/main/java/fredboat/command/music/control/SkipCommand.java
```diff
@@ -135,29 +209,29 @@ private void skipInRange(GuildPlayer player, TextChannel channel, Member invoker
...
+            if (result.getLeft()) {
                 channel.sendMessage(MessageFormat.format(I18n.get(guild).getString("skipSuccess"), 1, atc.getEffectiveTitle())).queue();
             }
         }
     }
 
     @Override
     public String help(Guild guild) {
-        String usage = "{0}{1} OR {0}{1} n OR {0}{1} n-m\n#";
+        String usage = "{0}{1} OR {0}{1} n OR {0}{1} n-m OR {0}{1} cd t OR {0}{1} n cd t OR {0}{1} n-m cd t\n#";
```
Frederikam: There's no use case for letting the user choose a cooldown themselves. This only adds to complexity of both the code and the user experience.

FredBoat/src/main/java/fredboat/command/music/control/SkipCommand.java
```diff
@@ -42,17 +42,49 @@
...
 
+    /**
+     * The keyword to change the {@link #skipCooldown}
+     */
+    private static final String COOLDOWN_CMD = "cd";
+
+    /**
+     * The default cooldown for calling the {@link #onInvoke} method in milliseconds.
+     */
+    private static final int DEFAULT_SKIP_COOLDOWN = 1000;
```
Frederikam: I'm personally thinking 500ms would be better. Let's not try to rate limit the skip command more than it is.

FredBoat/src/main/java/fredboat/command/music/control/SkipCommand.java
```diff
@@ -42,17 +42,49 @@
...
+
+    /**
+     * The latest time {@link #onInvoke} was called in milliseconds.
+     */
+    private long lastInvokeCall;
+
+    /**
+     * The cooldown for calling the {@link #onInvoke} method in milliseconds.
+     */
+    private long skipCooldown;
```
Frederikam: These variables are shared with *all* servers as there is only one instance of SkipCommand. You are using the cooldown on everyone.

FredBoat/src/main/java/fredboat/command/music/control/SkipCommand.java
drdgvhbh: Like a case where they can choose a cooldown but not skip?

Frederikam: Like, can you imagine a case where this would be actually useful to set?

FredBoat/src/main/java/fredboat/command/music/control/SkipCommand.java
Frederikam: That sounds like the best solution. Make sure to store the guild id as opposed to the Guild object, as  JDA's entities may become stale upon disconnect.

Frederikam:
The code looks good, but I just noticed that you are trying to merge to `master`. Please open a new pull request against `development`.

Sorry that I forgot to write this in the contribution guidelines.

PR #224, opened 2017-07-14
issue #189

drdgvhbh:
should be good now.


napstr:
This introduced a bug where the first skip after a restart of a the bot for a guild is ignored.
Will be fixed in one of my upcoming commits/PRs.

A log.info() or .debug() of ignored skips would have helped a lot figuring out what's happening, please don't forget to add those.

Issue #228, opened 2017-07-26
FredBoat does not ignore messages from other bots.


Vandesdelca32:
This does include command strings that another bot may send.

RealKC:
Seems fixable by checking if the invoker is a bot or not

Vandesdelca32:
Was just going to submit a pull request with the bit that needs a poke.

napstr:
Rip Skynet
#229

PR #229, opened 2017-07-26
Make FredBoat ignore messages from other bots

Vandesdelca32:
Get FredBoat to ignore messages sent by other bots.

I have not tested this code. But it should be all that is needed to stop Issue #228, unless it listens elsewhere as well.

### DIFF
--- a/FredBoat/src/main/java/fredboat/event/EventListenerBoat.java
+++ b/FredBoat/src/main/java/fredboat/event/EventListenerBoat.java
@@ -82,6 +82,10 @@ public class EventListenerBoat extends AbstractEventListener {
             return;
         }
 
+        if (event.getAuthor().isBot()) {
+            return;
+        }
+
         if (event.getAuthor().equals(event.getJDA().getSelfUser())) {
             log.info(event.getGuild().getName() + " \t " + event.getAuthor().getName() + " \t " + event.getMessage().getRawContent());
             return;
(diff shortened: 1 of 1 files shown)


RealKC:
@Vandesdelca32 I just noticed you are trying to merge into master, you need to merge into development(you may need to cherrypick the PR if when trying to merge into development commits from master appear).

Vandesdelca32:
Changed the target of the PR... It looks like it should still be okay to merge into this branch, only changes a couple of lines in /src/main/java/fredboat/event/EventListenerSelf.java

Shredder121:
>`event.getAuthor().equals(event.getJDA().getSelfUser())`
>`event.getAuthor().isBot()`

I think these are equivalent in FredBoat's case?
What do you think? @napstr

Vandesdelca32:
@Shredder121 The first appears to refer to the bot itself's message, where as the second is the case for any other bots that are posting, not just FredBoat.

Shredder121:
I thought we also dropped our own events, and since `isBot()` is true when it's the self user I figured we could collapse it.
But never mind.

Vandesdelca32:
I don't even know why that didn't cross my mind, sorry D:

napstr:
@Shredder121 they are, or rather, `event.getAuthor().equals(event.getJDA().getSelfUser())` will never be reached when `event.getAuthor().isBot()` already returns.

@Vandesdelca32 can you please move your addition below the statement where you placed it currently?

PR #233, opened 2017-07-29
Allow FredBoat to send ;;music in a DM

AndrielChaoti:
Just what the title says, this pull request will change it so that FredBoat can send the output of the `;;music` command to the user who ran it.

There could be some minor additional tweaks added to the message the bot sends, but it is tested and working.  Also #229


Shredder121:
Ah, I see you proposed a patch already.
What does this do when both the main, and the music bot listen for the help?
Do you then get 2 DMs?

Vandesdelca32:
The code changes left the basic structure of the code intact, so it should, in theory, respond the same... The output of the command will whisper you the music block and print the list of basic commands to the chat it was run in.

Vandesdelca32:
I went and checked this by adding the public bot to my testing/friend discord, and both bots will respond to both `;;help` and `;;commands`, but they're both running on the same prefix, so I feel this is to be expected.

FredBoat/src/main/java/fredboat/command/util/MusicHelpCommand.java
```diff
@@ -83,16 +86,33 @@ public void onInvoke(Guild guild, TextChannel channel, Member invoker, Message m
...
                 out = "";
             }
             out += s + "\n";
         }
-        channel.sendMessage(TextUtils.asMarkdown(out)).queue();
+        sendCommandsHelp(guild, channel, invoker, out);
+    }
+
+    public static void sendCommandsHelp(Guild guild, TextChannel channel, Member invoker, String dmMsg) {
+        // send the commands to the user in a DM
```
Shredder121: *Personal Preference*: if you name the method `sendCommandsHelpInDM` there's no need to pepper comments like these.

FredBoat/src/main/java/fredboat/command/util/MusicHelpCommand.java
```diff
@@ -83,16 +86,33 @@ public void onInvoke(Guild guild, TextChannel channel, Member invoker, Message m
...
+        invoker.getUser().openPrivateChannel().queue(privateChannel -> {
+            privateChannel.sendMessage(TextUtils.asMarkdown(dmMsg)).queue();
+            String out = I18n.get(guild).getString("helpSent"); //TODO: Replace this key with something better.
+            TextUtils.replyWithName(channel, invoker, out);
```
Shredder121: I think doing this in the `.queue` of the `.sendMessage` would allow for better reasoning when the help in fact *wasn't* sent, for example if the user blocks DMs from non-friends.
Otherwise it would report "sent the help in DMs" but it didn't actually happen.

FredBoat/src/main/java/fredboat/command/util/MusicHelpCommand.java
```diff
@@ -83,16 +86,33 @@ public void onInvoke(Guild guild, TextChannel channel, Member invoker, Message m
...
+        return musicComms;
+    }
+
+    public static void getFormattedCommandHelp(Guild guild, TextChannel channel, Member invoker) {
+        //retrieve the array of commands and messages.
+        final List<String> musicComms = getMusicComms(guild);
+
+        // Start building string:
         String out = "< " + I18n.get(guild).getString("helpMusicCommandsHeader") + " >\n";
+        // split the message into parts if it's too big pls.
```
Shredder121: *Personal Preference*: Comments should either 1) explain something that's not apparent from the code or 2) highlight some special contract.
That `getMusicComms` gets the music commands.. is apparent to me.
Thoughts?

FredBoat/src/main/java/fredboat/command/util/MusicHelpCommand.java
Vandesdelca32: I think this is actually a super good idea, especially from a UX perspective.

Vandesdelca32:
I tend to leave comments peppered in my code whether they're needed or not. They help me think when I'm writing, but don't necessarily have to stick around when I'm done.

FredBoat/src/main/java/fredboat/command/util/MusicHelpCommand.java
Vandesdelca32: Personal thoughts: I use comments more to position ideas and keep track of what's going to happen where while I'm writing... Like I'll start with
```java
private void someFunc() {
  // do the thing that we need to do
}
```
And sort of work from there.
*definately doesn't need to stay tho.* 😄

AndrielChaoti:
I can do that. Sorry about the CLA thing, this is the same account with a different username, and thus the CLA says I haven't signed it.

FredBoat/src/main/java/fredboat/command/util/HelpCommand.java
```diff
@@ -71,12 +71,14 @@ public String help(Guild guild) {
...
-            TextUtils.replyWithName(channel, invoker, out);
-        });
+        invoker.getUser().openPrivateChannel().queue( (privateChannel) ->
+                privateChannel.sendMessage(getHelpDmMsg(guild)).queue(message -> {
+                    // Let the user know when the message is successfully sent.
+                    String out = I18n.get(guild).getString("helpSent");
+                    out += "\n" + MessageFormat.format(I18n.get(guild).getString("helpCommandsPromotion"),
+                            "`" + Config.CONFIG.getPrefix() + "commands`");
+                    TextUtils.replyWithName(channel, invoker, out);
+        }));
```
Frederikam: We'll also need a consumer for failure. Just write a hardcoded string and I'll make sure it gets localized

FredBoat/src/main/java/fredboat/command/util/MusicHelpCommand.java
```diff
@@ -83,16 +86,33 @@ public void onInvoke(Guild guild, TextChannel channel, Member invoker, Message m
...
             out += s + "\n";
         }
-        channel.sendMessage(TextUtils.asMarkdown(out)).queue();
+        sendCommandsHelpInDM(guild, channel, invoker, out);
+    }
+
+    private static void sendCommandsHelpInDM(Guild guild, TextChannel channel, Member invoker, String dmMsg) {
+        invoker.getUser().openPrivateChannel().queue(privateChannel ->
+                privateChannel.sendMessage(TextUtils.asMarkdown(dmMsg)).queue( success -> {
+                    String out = "Commands Help has been sent to your DMs!"; // TODO: I18n
```
Frederikam: I feel that the other I18n string would be appropriate here

Frederikam: Eg the one we have for ;;help

AndrielChaoti:
Yep, I've made local builds and have it running on the Starbound official server. I just need someone that the bot can't actually DM, as I've not been able to find someone where that could happen :P

napstr:
> I just need someone that the bot can't actually DM, as I've not been able to find someone where that could happen :P

Setting Privacy Settings to disallow members of common servers to DM you will have that effect (need to do that for each shared server).

AndrielChaoti:
So I would have to do it for all the servers I am running the bot on. Being owner of the bot doesn't affect this then?

Kurozume:
@AndrielChaoti No if you do it on just one and do the command in that server it should work
Being bot owner does not affect it

_Sent from my Sony F8331 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #234, opened 2017-07-29
Hoist the lambda out of the loop

Shredder121:
Basically, for each user the lambda had to be bootstrapped and initialized.

In the after picture, the foreach just disappeared, because it didn't do any allocations anymore.

|  before   |  after
|---------|--------
| ![Before change][before] | ![After change][after]

[before]: https://user-images.githubusercontent.com/4105066/28748449-ac3a47de-74b8-11e7-8268-522047986ed3.png
[after]: https://user-images.githubusercontent.com/4105066/28748459-feac5bb0-74b8-11e7-8443-57a17b3bda7f.png

### DIFF
--- a/FredBoat/src/main/java/fredboat/feature/togglz/FeatureFlags.java
+++ b/FredBoat/src/main/java/fredboat/feature/togglz/FeatureFlags.java
@@ -49,11 +49,7 @@ public enum FeatureFlags implements Feature {
     @Label("Permissions")
     @EnabledByDefault
     PERMISSIONS,
-
-    //using data methods that don't collect everything to new data structures
-    @Label("Streaming data methods")
-    @EnabledByDefault
-    DATA_METHODS;
+    ;
 
     public boolean isActive() {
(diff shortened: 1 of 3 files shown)


FredBoat/src/main/java/fredboat/feature/togglz/FeatureFlags.java
```diff
@@ -49,11 +49,7 @@
     @Label("Permissions")
     @EnabledByDefault
     PERMISSIONS,
```
Shredder121: I also figured we could use a trailing comma in the feature flags.
Then the semicolon doesn't move around as much, so features shouldn't conflict anymore.
Hopefully ™️

Shredder121: (and also have the semicolon on its own line)

Shredder121:
Anyway, we're done with this one, I was mainly concerned if there are any other lambda issues, but we don't have any other hot path lambdas.

PR #235, opened 2017-07-30
Switch entity ids from String to long, clean up Entity Reader/Writer

napstr:
This will need some column alterations to be done when deploying.
Here are the necessary SQL queries:
ALTER TABLE public.guild_config ALTER COLUMN guildid TYPE bigint USING guildid::bigint
ALTER TABLE public.user_config ALTER COLUMN userid TYPE bigint USING userid::bigint
ALTER TABLE public.guild_permissions ALTER COLUMN id TYPE bigint USING id::bigint

The discordant methods in the Entity classes and using String ids when they are actually long snowflakes have been grinding my gears _long_ enough. I want to change that before it gets worse (it does with persistent tracklists and whatever the orchestrator will be).

The thing to keep in mind when applying this is to update the patron bot (and main) too.

### DIFF
--- a/FredBoat/src/main/java/fredboat/api/OAuthManager.java
+++ b/FredBoat/src/main/java/fredboat/api/OAuthManager.java
@@ -85,17 +85,17 @@ public class OAuthManager {
     private static UConfig saveTokenToConfig(TokenGrant token){
         User user = DiscordUtil.getUserFromBearer(FredBoat.getFirstJDA(), token.getBearer());
 
-        UConfig uconfig = EntityReader.getUConfig(user.getId());
+        UConfig uconfig = EntityReader.getEntity(user.getIdLong(), UConfig.class);
 
         uconfig = uconfig == null ? new UConfig() : uconfig;
 
         uconfig.setBearer(token.getBearer())
                 .setBearerExpiration(token.getExpirationTime())
                 .setRefresh(token.getRefresh())
-                .setUserId(user.getId());
+                .setUserId(user.getIdLong());
 
         //Save to database
-        EntityWriter.mergeUConfig(uconfig);
+        uconfig = EntityWriter.merge(uconfig);
 
         return uconfig;
     }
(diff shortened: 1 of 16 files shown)


Frederikam:
Does this have any side effects besides migrating the database for all FredBoat bots?

napstr:
No.
Might rustle some selfhosters but I don't consider that more important than having efficient and less complicated/error prone code.

Frederikam:
v1.1 it is then

Currently waiting to see if Shredder wants to review this, as I personally have other things to take care of today.

napstr:
v1.1 sounds good. As for our setup I was planning to get this  live in between deployment of major changes to keep things simple and revertable.

knight-ryu12:
Oh yea. that sounds good for me tho

FredBoat/src/main/java/fredboat/api/OAuthManager.java
```diff
@@ -85,17 +85,17 @@ public static UConfig ensureUnexpiredBearer(UConfig config) {
     private static UConfig saveTokenToConfig(TokenGrant token){
         User user = DiscordUtil.getUserFromBearer(FredBoat.getFirstJDA(), token.getBearer());
 
-        UConfig uconfig = EntityReader.getUConfig(user.getId());
+        UConfig uconfig = EntityReader.getEntity(user.getIdLong(), UConfig.class);
 
         uconfig = uconfig == null ? new UConfig() : uconfig;
```
Shredder121: Probably a product of earlier refactoring and NPE anxious code.
But the `.getEntity` now has a non-null returning contract:
```java
//return a fresh object if we didn't find the one we were looking for
if (config == null) config = newInstance(id, clazz);
return config;
```

napstr: Nice catch. UConfig isn't implemented at all. As soon as someone touches it he can clean up in there I guess.

napstr:
hbm2ddl was set to only add new columns, never touch old ones.

napstr:
What's the status here?

Frederikam:
We're just not going to merge this before we update the bot and all the DB schemas.

Issue #199, opened 2017-06-23
[;;repeat all] Known bugs


ObeIisk:
- `;;destroy` does not actually clear the queue (note sure if it even "resets" the player) when the list is on repeat
**`;;destroy` is supposed to clear the queue, no matter if repeated or not.**

- `;;repeat all` `;;skip n-m` will remove the songs from the list, but saying `;;skip` will keep them
**`;;skip` is supposed to remove the songs from the list.**

Shredder121:
> ;;skip is supposed to remove the songs from the list.

I understand the thought behind skip, but when I have a playlist on repeat, I want to be able to navigate inside the queue without losing tracks.
Should I open another issue for this?

Frederikam:
>I understand the thought behind skip, but when I have a playlist on repeat, I want to be able to navigate inside the queue without losing tracks.

You could do something like `;;fwd 50:00:00`

Nanabell:
> You could do something like ;;fwd 50:00:00

that is kinda meh as user experience

Frederikam:
It's not a common use case though

napstr:
Currently forwarding over the track length will do the same as skipping the track, however I'm adjusting it to be treated as not a skip (similar to how split tracks ending is handled), so that behavior will look like what Fred describes in this thread.

napstr:
This was done with the refactoring of the music stack that took place last week following the lavalink deploy.

PR #236, opened 2017-08-06
Refactoring of the music stack

napstr:
This cleans up everything weird going on in the music stack.

It builds on #235 

It should have been part of persistent track lists (and guild players), but I'm still working on their performance at scale.

### DIFF
--- a/FredBoat/src/main/java/fredboat/FredBoat.java
+++ b/FredBoat/src/main/java/fredboat/FredBoat.java
@@ -90,7 +90,8 @@ public abstract class FredBoat {
     ShardWatchdogListener shardWatchdogListener = null;
 
     //For when we need to join a revived shard with it's old GuildPlayers
-    final ArrayList<String> channelsToRejoin = new ArrayList<>();
+    protected boolean reviving = false;
+    protected final ArrayList<Long> channelsToRejoin = new ArrayList<>();
 
     //unlimited threads = http://i.imgur.com/H3b7H1S.gif
     //use this executor for various small async tasks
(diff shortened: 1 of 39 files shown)


Frederikam:
If you provide a list of changes you've made it would make this easier to review.

napstr:
Here's a list I've been keeping:

Fix bug introduced in #224 where the first skip in a guild after a restart would always be ignored
Fix issue #199
Fix missing checks where a user moving to a different channel would not trigger pausing, and a user moving into the channel from a different one would not trigger auto_resume, and FredBoat itself being moved around wouldn't react to it either.
Fix a bug where a bot user joining into the channel would trigger the autoresume
Do the skip announcement before the announcement of the next playing track.


Additionally:
More (efficient) methods have been defined for ITrackProvider and implemented in the SimpleTrackProvider. All shuffle and repeat related stuff is in the AbstractTrackProvider. GuildPlayer and AbstractPlayer are much more lean and methods have been given better names (there were like three playSomething methods including an infamous play0() one lol essentially doing very similar things). There was some confusion about which events happen when exactly and how (split tracks ending, skipping) and weird code written around to handle that which I stripped down. Errors loading a track (to play) would lead to no announcement and the player stopping, that's handled better now.

Frederikam:
Just skimmed this. I noticed how you changed the IDs of the entities to longs. Is that not duplicate of #235? And does that not also mean that this cannot be merged before we update all FredBoat instances and migrate all the schemas?

napstr:
Yes, that's why I opened #235 in advance, to not have it tangled up with other updates.
Tbh I probably found a different and better solution to it yesterday which would be to just generify IEntity and the id. That way we don't have to touch any existing tables. The desire to use longs instead of Strings as the main id comes from the Ratelimiter which holds a ton of objects in memory (I did some calculations back in the day...). The impact on other entities is marginal and switching them to longs is probably not worth the trouble if there is another way.

napstr:
closed in favor of #245

PR #243, opened 2017-08-11
Use the PermSystem for the ;;lang ... command

Nanabell:


Nanabell:
haha no this doesnt work

PR #249, opened 2017-08-21
Allow BOT_ADMINs to use "Destroy" regardless of permissions

### LINKED ISSUE #231: BOT_ADMIN not overriding manage messages on ;;destroy
More or less what the title says, BOT_ADMIN cannot use ;;destroy without having a role for manage messages, I don't know if the requirement should be lifted, or if that requirement should stay, but be bypassed for BOT_ADMIN

Nanabell:
image
image
I left the "destroyDenied" message untouched as it does not really affect anyone but is Bot Admins.   
This fixes #231

### DIFF
--- a/FredBoat/src/main/java/fredboat/command/music/control/DestroyCommand.java
+++ b/FredBoat/src/main/java/fredboat/command/music/control/DestroyCommand.java
@@ -44,7 +44,7 @@ public class DestroyCommand extends Command implements IMusicCommand, ICommandRe
     @Override
     public void onInvoke(Guild guild, TextChannel channel, Member invoker, Message message, String[] args) {
         if(invoker.hasPermission(channel, Permission.MESSAGE_MANAGE)
-                || PermsUtil.isUserBotOwner(invoker.getUser())) {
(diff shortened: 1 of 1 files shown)


Nanabell:
On second thought should "ADMINS" also be allowed to use it? The issue only mentions BOT_ADMINS so i didn't think about it first

napstr:
Admins of a discord server have the `Permission.MESSAGE_MANAGE`.

Nanabell:
I was thinking about our PermissionLevel Admins since you don't have to have Discord Admin perms to have Fredboat Admin Perms

napstr:
The destroy command can be used by anyone with `Permission.MESSAGE_MANAGE` (which includes discord  server admins) [first statement of the if clause] and the ~~bot owner~~ FredBoat admins (which includes the bot owner) [second statement of the clause, which is what gets changed in this PR].

Looks good to me.

Nanabell:
I mean refer to Fred's screenshot in #231. i meant the ;;admin add <id> Admin not Discord admin or BOT_ADMINS

napstr:
Might as well allow them too.

PR #250, opened 2017-08-22
Show the length of a song in the "list" command

### LINKED ISSUE #247: Make ;;queue also show track length
I think the title is pretty self-explanatory.

Nanabell:
image

ref: Frederikam/FredBoat-i13n#18  
fixes #247
depended on Frederikam/FredBoat-i13n#18


Shredder121:
*Idea*: merge the two monospace blocks, so the monospaced items are all aligned.
(Doesn't fix it when there is one track with a length measured in hours though.
Thoughts?

Nanabell:
I guess I take that as "the design was ok"

Frederikam:
Yep

PR #258, opened 2017-08-27
Advertise patreon on leave setting

Frederikam:
This adds a feature flag that allows FredBoat to advertise FredBoat Patron when ;;leave is called.

This only works under the following conditions:
- The feature is enabled (disabled by default, no selfhosting hassle)
- The player has existed for at least an hour
- The message has not been posted the last 2 hours for that guild/player

TODO: Translations

Thoughts?

### DIFF
--- a/FredBoat/src/main/java/fredboat/audio/GuildPlayer.java
+++ b/FredBoat/src/main/java/fredboat/audio/GuildPlayer.java
@@ -35,6 +35,7 @@ import fredboat.db.DatabaseNotReadyException;
 import fredboat.db.EntityReader;
 import fredboat.db.entity.GuildConfig;
 import fredboat.feature.I18n;
+import fredboat.feature.togglz.FeatureFlags;
 import fredboat.perms.PermissionLevel;
 import fredboat.perms.PermsUtil;
 import fredboat.util.TextUtils;
@@ -54,6 +55,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 public class GuildPlayer extends AbstractPlayer {
 
@@ -63,8 +65,9 @@ public class GuildPlayer extends AbstractPlayer {
     private final String guildId;
     public final Map<String, VideoSelection> selections = new HashMap<>();
     private String currentTCId;
-
     private final AudioLoader audioLoader;
+    private long earliestPotentialDonationNag = System.currentTimeMillis()
+            + TimeUnit.MINUTES.toMillis(60);
 
     @SuppressWarnings("LeakingThisInConstructor")
     public GuildPlayer(Guild guild) {
@@ -115,7 +118,21 @@ public class GuildPlayer extends AbstractPlayer {
             if (manager.getConnectedChannel() == null) {
                 channel.sendMessage(I18n.get(getGuild()).getString("playerNotInChannel")).queue();
             } else {
-                channel.sendMessage(MessageFormat.format(I18n.get(getGuild()).getString("playerLeftChannel"), getChannel().getName())).queue();
+                String msg = MessageFormat.format(I18n.get(getGuild()).getString("playerLeftChannel"), getChannel().getName());
+
+                // Note that we will only nag users if they seem to actually understand and use the bot
+                if (FeatureFlags.ADVERTISE_DONATION_ON_LEAVE.isActive()
+                        && getTotalRemainingMusicTimeSeconds() > 300
+                        && System.currentTimeMillis() > earliestPotentialDonationNag) {
+                    String translated = "If you find FredBoat useful, please consider donating so that we can keep the lights on.";
+                    msg = msg + "\n\n"
+                            + translated
+                            + "\n<https://fredboat.com/docs/donate>";
+
+                    earliestPotentialDonationNag = System.currentTimeMillis() + TimeUnit.MINUTES.toMillis(120);
+                }
+
+                channel.sendMessage(msg).queue();
             }
         }
         manager.closeAudioConnection();
--- a/FredBoat/src/main/java/fredboat/feature/togglz/FeatureFlags.java
+++ b/FredBoat/src/main/java/fredboat/feature/togglz/FeatureFlags.java
@@ -55,6 +55,9 @@ public enum FeatureFlags implements Feature {
 
     @Label("Force soundcloud search instead of youtube")
     FORCE_SOUNDCLOUD_SEARCH,
+
+    @Label("Advertise donation page on leave")
+    ADVERTISE_DONATION_ON_LEAVE,
     ;
 
     public boolean isActive() {


FredBoat/src/main/java/fredboat/audio/GuildPlayer.java
```diff
@@ -63,8 +65,9 @@
     private final String guildId;
     public final Map<String, VideoSelection> selections = new HashMap<>();
     private String currentTCId;
-
     private final AudioLoader audioLoader;
+    private long earliestPotentialDonationNag = System.currentTimeMillis()
+            + TimeUnit.MINUTES.toMillis(60);
```
Shredder121:
```java
private LocalDateTime earliestPotentialDonationNag = LocalDateTime.now().plusHours(1);
```

FredBoat/src/main/java/fredboat/audio/GuildPlayer.java
```diff
@@ -115,7 +118,21 @@ public void leaveVoiceChannelRequest(TextChannel channel, boolean silent) {
             if (manager.getConnectedChannel() == null) {
                 channel.sendMessage(I18n.get(getGuild()).getString("playerNotInChannel")).queue();
             } else {
-                channel.sendMessage(MessageFormat.format(I18n.get(getGuild()).getString("playerLeftChannel"), getChannel().getName())).queue();
+                String msg = MessageFormat.format(I18n.get(getGuild()).getString("playerLeftChannel"), getChannel().getName());
+
+                // Note that we will only nag users if they seem to actually understand and use the bot
+                if (FeatureFlags.ADVERTISE_DONATION_ON_LEAVE.isActive()
+                        && getTotalRemainingMusicTimeSeconds() > 300
+                        && System.currentTimeMillis() > earliestPotentialDonationNag) {
```
Shredder121:
```java
&& LocalDateTime.now().isAfter(earliestPotentialDonationNag)
```

FredBoat/src/main/java/fredboat/audio/GuildPlayer.java
```diff
@@ -115,7 +118,21 @@ public void leaveVoiceChannelRequest(TextChannel channel, boolean silent) {
...
+                    String translated = "If you find FredBoat useful, please consider donating so that we can keep the lights on.";
+                    msg = msg + "\n\n"
+                            + translated
+                            + "\n<https://fredboat.com/docs/donate>";
+
+                    earliestPotentialDonationNag = System.currentTimeMillis() + TimeUnit.MINUTES.toMillis(120);
```
Shredder121:
```java
earliestPotentialDonationNag = LocalDateTime.now().plusHours(2);
```

Shredder121:
>This only works under the following conditions:

Also if the remaining time in the queue is 300 seconds yeah?

Frederikam:
That too

Frederikam:
Thinking I want this to depend on the actual time a server has played audio

Frederikam:
Only thing needed now would be the translation. I'm worried about the performance of this, as we would have up to several thousands of tasks as GuildPlayers starts to pile up. The task is only executed every 10 minutes though. Is there a way we can improve this?

FredBoat/src/main/java/fredboat/audio/PlaybackTimeMonitor.java
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Monitors playback time
+ */
+public class PlaybackTimeMonitor {
+
+    private static final Logger log = LoggerFactory.getLogger(PlaybackTimeMonitor.class);
+
+    private volatile Duration playbackTime = Duration.ZERO;
+    private final ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor();
```
napstr: A full ExecutorService for each Object of this sounds like an overkill.

FredBoat/src/main/java/fredboat/audio/PlaybackTimeMonitor.java
```diff
@@ -0,0 +1,71 @@
...
+
+    PlaybackTimeMonitor(GuildPlayer guildPlayer) {
```
napstr: Is holding a reference to a GuildPlayer a good idea? ;;destroy will walk right over it.

FredBoat/src/main/java/fredboat/audio/PlaybackTimeMonitor.java
Frederikam: I'm not familiar with pooling threads. How would you do it?

FredBoat/src/main/java/fredboat/audio/PlaybackTimeMonitor.java
Frederikam: But wouldn't the `PlaybackTimeMonitor` be finalized then? The intention was for the `PlaybackTimeMonitor` to have the same lifespan as the `GuildPlayer`.

FredBoat/src/main/java/fredboat/audio/PlaybackTimeMonitor.java
napstr: I would use a static ScheduledExecutorService with a single thread. The tasks aren't doing any heavy lifting.

FredBoat/src/main/java/fredboat/audio/PlaybackTimeMonitor.java
napstr: There is no guarantee anything will be finalized in the JVM [(random SO thread which seems to be a good read on the topic)](https://stackoverflow.com/questions/2506488/when-is-the-finalize-method-called-in-java).

Having said that, I'm not 100% sure, but it would be close enough for me personally to avoid it, that the scheduled task holding a reference to the guild player might be a GC root and cause leaks. I think a possible solution could be to cancel the task when the GuildPlayer is destroyed.

Shredder121: And I can assure you, there are more than enough GC (and thus also finalizer) runs in the lifetime of the VM in FredBoat.
So unless the finalizer thread is killed by a bad finalizer, we can rely on it.

The popular opinion is only "don't rely *just* on the finalizer".
For example `File(In/Out)putStream` has a finalizer, but it's better to `close()` it explicitly.

napstr:
Just a note on this one, the join/leave stuff is a bit wonky right now after introducing lavalink. I'll try drilling down on it so we can merge this.

Frederikam:
Please put this on hold for now. I want to rework how leaving/joining works with lavalink

napstr:
Ok. Here's what I found so far: 
We are currently using two different ways in different places to get our current VoiceChannel;
```java
guild.getSelfMember().getVoiceState().getChannel();
```

```java
LavalinkManager.ins.getConnectedChannel(guild);
```
and the latter one is not being updated. The VoiceServerUpdateInterceptor seems to miss updates to the voice channel when users move FredBoat around. Which is not a complete disaster as the music keeps playing in the new channel after moving FredBoat, but it fucks with our commands and auto resume / auto pause logic.

napstr:
Should probably also mention the patron bot when ppl use the volume command.

Issue #259, opened 2017-08-27
Bombard youtube with less search queries (Spotify main offender)


napstr:
Since we turned off cookies due to a lot of 503s happening on Youtube's side, Youtube started flagging us as a suspicious and presenting a captcha for searches occasionally.
One of the main offenders is the Spotify playlist loader, which searches for each song's name and artist from a Spotify playlist, resulting in hundreds of search queries at the same time.

We could try rate limiting the searches, which would lead to an initially longer loading time for such a playlist, while at the same time caching the search results. The latter can be a thing for all searches, not just the ones done when loading a Spotify playlist, and should speed up things overall at the cost of some (?) memory.

To discuss: What would be a good caching duration? What's an acceptable memory cost?

#### Steps to resolve
updated 28th august

- [x] ~~Rate limit searches when loading a Spotify playlist~~ This did not help much, Youtube still flags us at a rate of 1 search per second, which is ridiculously slow already. Instead: Have a single thread load the Spotify playlists.
- [x] Cache all searches on a search provider scope (only successful ones!) #261
- [x] Patron only: Use Youtube API if the lavaplayer search returns a 503 #261
- [x] Fallback to Soundcloud search if none of the above provided a result #261

Frederikam:
Why memory? Why not persist it to the DB?

napstr:
Sure, that's a great option. I wanted to try Guava's Cache for this. I wonder if it supports caching into a database out of the box.

Frederikam:
There's always EHCache as the second level cache

napstr:
Nope, Guava isn't meant for the db, just memory.
![image](https://user-images.githubusercontent.com/6048348/29750331-2299aa5a-8b3e-11e7-9232-9f0d41287d51.png)

https://github.com/google/guava/wiki/CachesExplained#applicability

Frederikam:
But we *do* have EGCache already hooked up to Hibernate for entity caching

napstr:
I don't think it's a big issue, the main difference to using Guava will be handling the expiration.

Frederikam:
We could also use ehcache to persist directly to the FS without the DB, with an expiration

napstr:
Ratelimiting the searches is a bit tricky. When I loaded a 1488 track playlist with the current, unlimited code, it got 900 something results back before being flagged by Youtube.
Then I put a rate limit in place that fires a search request only once per second, and ended up getting flagged after 100 something results.

napstr:
[...]

PR #261, opened 2017-08-27
Cache search results + fallback to Youtube API

napstr:
This will cache all searches happening through SearchUtil for 24 hours, reducing our calls to Youtube (and Soundcloud).

Not only does it reduce our calls to Youtube, it speeds up the whole execution of the command (subjective impression here).

Part of the solution of #259

### DIFF
--- a/FredBoat/src/main/java/fredboat/db/entity/SearchResult.java
+++ b/FredBoat/src/main/java/fredboat/db/entity/SearchResult.java
@@ -35,6 +35,7 @@ import fredboat.FredBoat;
 import fredboat.db.DatabaseManager;
 import fredboat.db.DatabaseNotReadyException;
 import fredboat.util.rest.SearchUtil;
+import org.apache.commons.lang3.SerializationUtils;
 import org.hibernate.annotations.Cache;
 import org.hibernate.annotations.CacheConcurrencyStrategy;
 import org.slf4j.Logger;
@@ -79,7 +80,7 @@ public class SearchResult {
 
     @Lob
     @Column(name = "search_result")
-    private SerializableAudioPlaylist encodedSearchResult;
+    private byte[] serializedSearchResult;
 
     //for JPA
     public SearchResult() {
@@ -89,7 +90,7 @@ public class SearchResult {
                         AudioPlaylist searchResult) {
         this.searchResultId = new SearchResultId(provider, searchTerm);
         this.timestamp = System.currentTimeMillis();
-        this.encodedSearchResult = new SerializableAudioPlaylist(playerManager, searchResult);
+        this.serializedSearchResult = SerializationUtils.serialize(new SerializableAudioPlaylist(playerManager, searchResult));
     }
 
     /**
@@ -180,11 +181,12 @@ public class SearchResult {
     }
 
     public AudioPlaylist getSearchResult(AudioPlayerManager playerManager) {
-        return encodedSearchResult.decode(playerManager);
+        SerializableAudioPlaylist sap = SerializationUtils.deserialize(serializedSearchResult);
+        return sap.decode(playerManager);
     }
 
     public void setSearchResult(AudioPlayerManager playerManager, AudioPlaylist searchResult) {
-        this.encodedSearchResult = new SerializableAudioPlaylist(playerManager, searchResult);
+        this.serializedSearchResult = SerializationUtils.serialize(new SerializableAudioPlaylist(playerManager, searchResult));
     }
 
     /**
--- a/FredBoat/src/main/java/fredboat/db/entity/SearchResult.java
+++ b/FredBoat/src/main/java/fredboat/db/entity/SearchResult.java
@@ -1,0 +1,334 @@
+/*
+ *
+ * MIT License
+ *
+ * Copyright (c) 2017 Frederik Ar. Mikkelsen
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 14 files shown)


FredBoat/src/main/java/fredboat/db/entity/SearchResult.java
```diff
@@ -0,0 +1,334 @@
...
+
+/**
+ * Created by napster on 27.08.17.
+ * <p>
+ * Caches a search result
+ */
+@Entity
+@Table(name = "search_results")
+//todo does this need to be added to ehcache.xml?
+@Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE, region = "search_results")
```
napstr: @Frederikam's is this annotation enough for ehcache to work, or do we need an entry in ehcache.xml additionally or something else?

Shredder121: Pretty sure you.. also need to mention it in ehcache.xml... :/
The last commit of the ehcache.xml was also for a new cache entry.
https://github.com/Frederikam/FredBoat/commit/cbd7cd949ecb3a4ed6fdfc9de0a2d323c7620eea

Frederikam: Yeah you need to specify a region

napstr:
Reminder to myself: Test performance with a remote database and SQLite...I'm sensing an upcoming Vietnam flashback regarding that...

napstr:
This is ready to be reviewed and pulled.

FredBoat/src/main/java/fredboat/util/rest/SearchUtil.java
```diff
@@ -53,12 +70,86 @@ private static AudioPlayerManager initPlayerManager() {
...
+                return cacheResult;
+            }
+
+            //2. lavaplayer
+            try {
+                AudioPlaylist lavaplayerResult = new SearchResultHandler().searchSync(provider, query, timeoutMillis);
+                if (!lavaplayerResult.getTracks().isEmpty()) {
+                    log.debug("Loaded search result {} {} from lavaplayer", provider, query);
+                    // got a search result? cache and return it
+                    FredBoat.executor.submit(() -> new SearchResult(PLAYER_MANAGER, provider, query, lavaplayerResult).save());
```
Shredder121: `submit` without using the return value
It's best to use `execute` then

FredBoat/src/main/java/fredboat/util/rest/SearchUtil.java
```diff
@@ -53,12 +70,86 @@ private static AudioPlayerManager initPlayerManager() {
...
+
+            //3. optional: youtube api
+            if (provider == SearchProvider.YOUTUBE &&
+                    (Config.CONFIG.isPatronDistribution() || Config.CONFIG.isDevDistribution())) {
+                try {
+                    AudioPlaylist youtubeApiResult = YoutubeAPI.search(query, 5, PLAYER_MANAGER.source(YoutubeAudioSourceManager.class));
+                    if (!youtubeApiResult.getTracks().isEmpty()) {
+                        log.debug("Loaded search result {} {} from Youtube API", provider, query);
+                        // got a search result? cache and return it
+                        FredBoat.executor.submit(() -> new SearchResult(PLAYER_MANAGER, provider, query, youtubeApiResult).save());
```
Shredder121: Same here, `execute`

FredBoat/src/main/java/fredboat/util/rest/SearchUtil.java
```diff
@@ -53,12 +70,86 @@ private static AudioPlayerManager initPlayerManager() {
...
+     * @param providers     Providers that shall be used for the search. They will be used in the order they are provided, the
+     *                      result of the first successful one will be returned
+     * @return The result of the search, or an empty list.
+     * @throws SearchingException If none of the search providers could give us a result, and there was at least one SearchingException thrown by them
+     */
+    public static AudioPlaylist searchForTracks(String query, long cacheMaxAge, int timeoutMillis, List<SearchProvider> providers)
+            throws SearchingException {
+
+        List<SearchProvider> provs = new ArrayList<>();
+        if (providers == null || providers.size() == 0) {
```
Shredder121: There is an `isEmpty()` method :C

FredBoat/src/main/java/fredboat/util/rest/SearchUtil.java
```diff
@@ -53,12 +70,86 @@ private static AudioPlayerManager initPlayerManager() {
...
+
+        SearchingException exception = null;
+
+        for (SearchProvider provider : provs) {
+            //1. cache
+            AudioPlaylist cacheResult = fromCache(provider, query, cacheMaxAge);
+            if (cacheResult != null && !cacheResult.getTracks().isEmpty()) {
+                log.debug("Loaded search result {} {} from cache", provider, query);
+                return cacheResult;
+            }
```
Shredder121: What if the youtube site was hit, but got 503?
Will an empty tracklist be persisted?
So will it always retry then? 🤔

napstr: If a 503 is returned, it will use the official Youtube API / Soundcloud, and that result will be cached and returned for future searches. The regular ;;play command that is, if a user used the ;;yt command, which only searches Youtube and nothing else (introduced in this PR, similar to the existing ;;sc command for soundcloud), then yes, it would be hitting the 503 constantly.

I'm thinking about detecting the 503 and putting the youtube search on a cooldown of a few minutes. Maybe that will help come back from the flagged status faster, and also improve response times as it's one (highly likely to fail) trip less to be made.

PR #264, opened 2017-08-30
Use the permission system for config command

Nanabell:
Another one of those pebble prs.
I don't know if this is intended that config command is limited to Admins of a server only but by moving this to the perm system it's basically the same with the exception that Bot admins can also change these settings  
EDIT:
should we remove the now not used string from i13n ?


Frederikam:
Looks good. Can you also remove the `configNotAdmin` from the I18n repo?

PR #270, opened 2017-09-01
Make unsafe lazy initialization safe again

Shredder121:
This uses the holder class pattern, so the JVM can do the locking when initializing for us.

### DIFF
--- a/FredBoat/src/main/java/fredboat/util/rest/SpotifyAPIWrapper.java
+++ b/FredBoat/src/main/java/fredboat/util/rest/SpotifyAPIWrapper.java
@@ -58,7 +58,6 @@ public class SpotifyAPIWrapper {
     private static final String URL_SPOTIFY_AUTHENTICATION_HOST = "https://accounts.spotify.com";
 
     private static final org.slf4j.Logger log = LoggerFactory.getLogger(SpotifyAPIWrapper.class);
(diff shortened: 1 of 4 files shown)


napstr:
Could you add this pattern for the SpotifyAPIWrapper class too please?

napstr:
This is a nice pattern, the only caveat is that there must absolutely be no unhandled exception thrown during initialization.

One such thing could be happening, when the Ratelimiter constructor calls the constructor of the Blacklist, which in turn queries the database, which could be a source of rare but occasionally happening DatabaseNotReadyExceptions.

Shredder121:
Good.. point, yeah.
So maybe just use double checked locking and storing it manually then?

napstr:
Yep.

FredBoat/src/main/java/fredboat/util/rest/SpotifyAPIWrapper.java
```diff
@@ -58,7 +58,6 @@
     private static final String URL_SPOTIFY_AUTHENTICATION_HOST = "https://accounts.spotify.com";
 
     private static final org.slf4j.Logger log = LoggerFactory.getLogger(SpotifyAPIWrapper.class);
-    private static volatile SpotifyAPIWrapper SPOTIFYAPIWRAPPER;
```
Shredder121: @napstr I also removed this now.

PR #273, opened 2017-09-04
Fix ehcache config

napstr:
Restrict search results cache region
Reformat and update ehcache.xml to persistence strategy
Don't persist cache to disk when the database is local
Add missing JPA annotations and Serializable implementation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -135,3 +135,4 @@ gc-*.log
 *dbconf.yaml
 /config.yaml
 *fredboat.db
+*.ppk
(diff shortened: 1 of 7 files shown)


Frederikam:
I doubt you know less about this than me

napstr:
Read up a few things.

The overall strategy is to only use the memory cache when the database is local, remote FredBoats will use the their disk additionally.
Other than that tuned down the out-of-control search cache, and fixed some formalities which might or might not have prevented us using the cache properly so far.

napstr:
Maybe @Shredder121 has some experience with 2nd level caching to share? Other than that my suggestions is to merge this somewhen before updating tomorrow.

Issue #278, opened 2017-09-07
Change command name handling to no longer be case-sensitive


AndrielChaoti:
Currently, entering in any command in the wrong case, for example `;;Play` will just fail silently in the channel.

It would lend to a better UX if commands were not case-sensitive. (Only the names checking would need to be changed for this, not parsing arguments, as there is reason for those to be case-sensitive.)

Macley-Kun:
I think this one can be closed, seeing that the commitment has taken place successfully :)

PR #280, opened 2017-09-07
Only attempt to get lowercase command names.

AndrielChaoti:
This is an overly simple fix for #278

There may be some instances I am missing for this to fully work, and I have not gotten a chance to test this in a live boat just yet. I will update the PR when I've done so.

### DIFF
--- a/FredBoat/src/main/java/fredboat/commandmeta/CommandRegistry.java
+++ b/FredBoat/src/main/java/fredboat/commandmeta/CommandRegistry.java
@@ -39,10 +39,11 @@ public class CommandRegistry {
     private static HashMap<String, CommandEntry> registry = new HashMap<>();
 
     public static void registerCommand(String name, Command command, String... aliases) {
+        name = name.toLowerCase();
         CommandEntry entry = new CommandEntry(command, name);
         registry.put(name, entry);
         for (String alias : aliases) {
-            registry.put(alias, entry);
+            registry.put(alias.toLowerCase(), entry);
         }
     }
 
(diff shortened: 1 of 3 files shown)


Kurozume:
What would happen however in the unlikely event that either we decidd we need a command with an uppercase character or a selfhoster tries to add one?
In its current state this update would mean the bot would never reach aliases with uppercase characters in them

_Sent from my Xperia XZ using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

AndrielChaoti:
True enough. Perhaps having the command registry add command names and aliases in lower case as well?

If it came to the need for commands with uppercase names, then I suppose this would be moot. :)

AndrielChaoti:
I've given this a test on a discord server at this point. I feel that adding the `.toLowerCase` to the command registration system is also necessary before the PR should be merged.

FredBoat/src/main/java/fredboat/commandmeta/CommandRegistry.java
```diff
@@ -81,7 +82,7 @@ public String help(Guild guild) {
 
         CommandEntry(Command command, String name) {
             this.command = command;
-            this.name = name;
+            this.name = name.toLowerCase();
```
Shredder121: At this point the `name` had been lowercased already in `registerCommand`

AndrielChaoti: so it has... fixed that :smile:

Shredder121:
I'm not really a fan of `toLowerCase` without a `Locale` argument, as that usually fails the ['Turkey test'][turkey-test]
Now this is not really a big problem with server architecture programs, but since the commands are English alphabet only, is it an idea to add a `Locale.ENLGISH`?

Oh okay lol nvm.
[turkey-test]: https://stackoverflow.com/questions/796986/what-is-the-turkey-test

PR #283, opened 2017-09-08
Add Vote Skip Command

### LINKED ISSUE #26: Voteskipping
Command that skips once a majority has been reached

Nanabell:
### Don't Merge yet as the whole system is just proof of concept currently.

I've used the 500ms cooldown just like the normal `skip` command but I didn't look at rate-limiting it 

This PR is currently just to track progress & get feedback
- Get Feedback
- As I'm not very familiar with java I'm sorry for any mistakes. Please tell what is retardedly done 😄 
- Should this be a `;;config` option to decide % or disable completely?
- Or rather a feature flag?
- Both?
- Rate-limits?

DEFINITE TODOS B4 MERGE:
- [x] - I18n

- [x] - Test the system with multiple users in vc (#FeelsLonelyMan)
while I can change code numbers around I can't summon more real users to me so I would need some Test subjects 

fixes #26 (at some point)

### DIFF
--- a/FredBoat/src/main/java/fredboat/audio/player/AbstractPlayer.java
+++ b/FredBoat/src/main/java/fredboat/audio/player/AbstractPlayer.java
@@ -257,10 +257,6 @@ public abstract class AbstractPlayer extends AudioEventAdapterWrapped implements
     public void onTrackEnd(AudioPlayer player, AudioTrack track, AudioTrackEndReason endReason) {
         log.debug("onTrackEnd({} {} {}) called", track.getInfo().title, endReason.name(), endReason.mayStartNext);
 
-
-        //Clear all the VoteSkips from the Map so we can start anew.
-        VoteSkipCommand.guildSkipVotes.clear();
-
         if (endReason == AudioTrackEndReason.FINISHED || endReason == AudioTrackEndReason.STOPPED) {
             loadAndPlay();
         } else if(endReason == AudioTrackEndReason.CLEANUP) {
--- a/FredBoat/src/main/java/fredboat/audio/player/AbstractPlayer.java
+++ b/FredBoat/src/main/java/fredboat/audio/player/AbstractPlayer.java
@@ -48,6 +48,7 @@ import fredboat.audio.queue.SplitAudioTrackContext;
 import fredboat.audio.queue.TrackEndMarkerHandler;
 import fredboat.audio.source.PlaylistImportSourceManager;
 import fredboat.audio.source.SpotifyPlaylistSourceManager;
+import fredboat.command.music.control.VoteSkipCommand;
 import fredboat.commandmeta.MessagingException;
 import fredboat.shared.constant.DistributionEnum;
 import lavalink.client.player.IPlayer;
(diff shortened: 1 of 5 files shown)


FredBoat/src/main/java/fredboat/audio/player/AbstractPlayer.java
```diff
@@ -254,6 +255,10 @@ public static AudioPlayerManager getPlayerManager() {
     public void onTrackEnd(AudioPlayer player, AudioTrack track, AudioTrackEndReason endReason) {
         log.debug("onTrackEnd({} {} {}) called", track.getInfo().title, endReason.name(), endReason.mayStartNext);
 
+
+        //Clear all the VoteSkips from the Map so we can start anew.
+        VoteSkipCommand.guildSkipVotes.clear();
```
napstr: isn't this clearing it for all guilds?

Nanabell: yes.. yes it is 👀 whoops

napstr:
If this is going to be a standalone commands, instead of changing the default behaviour of the skip command, you probably want to add a hint for user to use voteskip when they fail to regular skip a track.

napstr:
One thing to test would be whether `onTrackEnd` is _really_ always called when the player stops playing. Like is it called when destroy is used?

Nanabell:
>If this is going to be a standalone command (instead of changing the default behaviour of the skip command) you probably want to add a hint for users to use voteskip when they fail to regular skip a track.

well that is what I'm asking what is wanted, should this stay as a standalone command and always be enabled or togglable see op

napstr:
from the #26 thread:
> The idea is to create a config option to enforce voteskip, while allowing admins to forcefully skip.

Nanabell:
should we make it a feature_flag as well in case we need to disable it for w/e reason ?
should it have the same rate limits as the normal skip command?

I'm asking these now as by the creation of #26 none of these things were a thing

napstr:
Feel free to add a Feature Flag. I personally don't think this feature warrants one but there are no real downsides to having them around. The real question is which skip system should be the default one ;-)

Keep the ratelimit, if its a single user listening (single vote = skip) it still applies and they should be skipping with a skip range instead.

Nanabell:
I am ready from my side. I will add the I13n repo commit once it's been pulled

Nanabell:
As a side note, I didn't add a config option for now as I don't think many will want to disable the vote skip completely but allow them to queue songs. if it becomes a popular request we can always add it

FredBoat/src/main/java/fredboat/command/music/control/VoteSkipCommand.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    private boolean hasVoted(Guild guild, Member member) {
+        List<Long> voters = guildSkipVotes.get(guild.getIdLong());
+        return voters.contains(member.getUser().getIdLong());
+    }
+
+    private void displayVoteList(CommandContext context) {
+        EmbedBuilder embed = CentralMessaging.getClearThreadLocalEmbedBuilder();
+        embed.setTitle("WIP");
```
napstr: Is this really done?

Nanabell: Read up. this will never be called

Nanabell: https://github.com/Frederikam/FredBoat/pull/283/files/0be0342e4c0428f1a5f8de92d713974675a49b96#diff-7b1e8cef1ea9c97440baf3081a7aecc5R74 I left it in in case we want to add a list in the future but I don't see a solid reason to add it right now but if users start requesting it it's a headstart

PR #288, opened 2017-09-11
Check FBH hardcoeded setting b4 perm check

Nanabell:
I didn't test it but this is just one check b4 another so I'm 99.99999% sure nothing can blow up because of this. it's simply doing the fbh hard-coded channel/owner check b4 the perm check as we have a special message which deletes itself while the perm check doesn't


napstr:
:ok_hand:

PR #289, opened 2017-09-11
Centralized Messaging

napstr:
#### Content:

- All messaging happens at one place: CentralMessaging
- Introduce a context pattern - the context provides all necessary information
  about the source and invoker of a command in a way safe from revives + convenience
  methods to interact with the context.
- All messaging is queued() by default, but can be waited on if absolutely necessary
- Reduce creation of new MessageBuilders and EmbedBuilder objects to the minimum necessary
- Delete deprecated selfbot code.
- Add i18n context convenience method

#### Reasons:
- Less clutter:
image
- Central messaging allows us to check all content that we send out (prevent `@everyone` abuse for example)
- I think the context pattern is dope af, may even be evolved into a fully fledged messaging system
- Easy metrics collection :upside_down_face:


napstr:
This is a _refactoring_ PR - don't be intimidated by the line counts, the behaviour is unchanged.

Shredder121:
what do we use the jetbrains annotations for?
if it's nullable and friends, can we use jsr305 instead?

napstr:
Nullable and friends. I don't really have preferences for the underlying annotation package, don't even know the proper rules, just wanted the code to look like code written by a real dev:tm:

Frederikam:
Always wanted to get rid of the selfbot code

PR #292, opened 2017-09-11
Add Prometheus metrics

napstr:
~~This is **NOT** production safe yet, because of unknown performance impact
I'm actually not sure how to test the performance impacts this might bring.~~

The Prometheus docs warn from using too many labels on a metric, and I did set up a few labels with high cardinality (command class), though these are one-dimensional, do it's _probably_ fine.


Shredder121:
Anyway for us to test, and see the performance impact ourselves? 🤔

napstr:
Call localhost:1356/metrics every 10 seconds

napstr:
Ok this boils down to "How to measure the impact of turning on instrumentation?", a truely enjoyable conundrum.

So to do this in a super clean and probably overly careful way, here's the plan:
Two (temporary) feature flags:
- Basic instrumentation
- Full instrumentation

Turn on the first, leave for a few days, turn on the second. Then we'll go from there.

napstr:
Scratch that...we should be good incrementing a few simple counters.
Added a feature flag for the crazy stats (multidimensional per command stats and timings), which we can enable later.

This is ready to be reviewed/merged.

napstr:
Want add that these have been running in some private 5 server boat of mine for a while without issues.

napstr:
About to merge this into development soon:tm:

Here's some more general info: Counters and Gauges are pretty cheap in our client so we are fine having lots of those, even with a few labels.
The only thing hidden behind a feature flag is the command execution time Histogram because its performance implications are unclear

This PR also features more thorough failure logging of JDA actions, and a small refactoring of the PlayerRegistry towards a singleton pattern.

PR #302, opened 2017-09-20
Handle Insufficent permission exceptions in messaging

napstr:
~~If we encounter missing permissions during anything messaging related, we will send a message saying which permission we are missing (for example embeds, or attaching files, etc).~~

~~If we encounter missing permissions while sending a message, we will just log that to prevent a loop. (this shouldn't be happening in the first place since we don't accept commands from channels that don't allow us to write messages).~~

~~Need feedback from @Frederikam whether a) this is an acceptable flow of things and b) whether i18n is needed, and if needed whether it can be postponed (which I would prefer; a commit doing all those missing i18n todos will happen in the near future anyways).~~

Implements the behaviour of the NowplayingCommand (telling the user when there are missing permissions to display embeds) in all messaging related methods for all potentially missing permissions.


napstr:
Don't pull yet, another commit and update of OP incoming.

napstr:
Ok, ready to be reviewed/merged.

PR #310, opened 2017-09-25
Let the end user know that there are tabs in their yaml files

Shredder121:
Napster and I discussed some things about the yaml files, and potentially we could also update the selfhost documentation to mention tabs are illegal in a yaml file.
So, thoughts? Should we do that instead of trying a fix-up?

### DIFF
--- a/FredBoat/config.yaml
+++ b/FredBoat/config.yaml
@@ -8,7 +8,7 @@ useAutoBlacklist:  true        # Set to true to automatically blacklist users wh
 game:              ""          # Set the displayed game/status. Leave empty quote marks for the default status
 
 enableYouTube:     true        # Set to true to enable playing YouTube links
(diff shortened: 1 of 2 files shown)


napstr:
I don't mind just replacing the tabs with spaces and starting. Make it as convenient as possible imo.

Shredder121:
Makes sense.
In general we're using 2 space indents, we could replace them with that instead of removing them?
I'll fiddle around a little to see how fluid that is.

Shredder121:
So 2 spaces works great, if it's used for indentation accidentally it lines up again, and if accidentally used for alignment there is no issue, as you don't need to have anything aligned in yaml.

Do we save the fixup, or only support non-tabs?

napstr:
Don't really have a strong opinion on that. I think keeping a warning is good enough. Those who care will adjust their files, those who don't won't.

PR #311, opened 2017-09-25
Fix split tracks. Relies on Frederikam/Lavalink#13

napstr:
Todo:

- [x] Update Lavalink to include this PR Frederikam/Lavalink#13

### DIFF
--- a/FredBoat/src/main/java/fredboat/audio/queue/SplitAudioTrackContext.java
+++ b/FredBoat/src/main/java/fredboat/audio/queue/SplitAudioTrackContext.java
@@ -26,6 +26,7 @@
 package fredboat.audio.queue;
 
 import com.sedmelluq.discord.lavaplayer.track.AudioTrack;
+import lavalink.client.player.TrackData;
 import net.dv8tion.jda.core.entities.Member;
 
 // TODO: Fix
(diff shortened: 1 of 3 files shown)


napstr:
Build will be automagically fixed when Lavalink is updated.

Frederikam:
Lavalink PR was merged, just need to change the version here

napstr:
How are we doing that currently? I was under the impression you would just bump a lavalink number.

Frederikam:
Hello Twitch!

Frederikam:
Just change the commit hash in the FredBoat POM

napstr:
Aight, forgot we're using jitpack, as I'm running local snapshot builds.

FredBoat/src/main/java/fredboat/audio/queue/SplitAudioTrackContext.java
```diff
@@ -45,6 +46,11 @@ public SplitAudioTrackContext(AudioTrack at, long userId, long guildId, long sta
         this.startPos = startPos;
         this.endPos = endPos;
         this.title = title;
+
+
+        // TODO introduce TrackData to FredBoat instead of the AudioTrackContext and SplitAudioTrackContext?
+        // TODO doing that would mean Yet Another Rewrite:tm: of the persistent tracklists
```
Frederikam: Not really necessary. AudioTrackContext is more flexible IMO

napstr: Indeed. Also feels like spilling FredBoat implementation details into Lavalink. Should have removed that comment as I did with a similar one. Will do in a sec.

Frederikam: Also remember to remove `@Deprecated`

PR #312, opened 2017-09-25
Use thread safe collection classes

Shredder121:
So, in Sentry there were still some ConcurrentModificationExceptions that reminded me to fix this.

Cleanup of wrapping some return values in `synchronized/unmodifiable` counterparts still needed.


napstr:
I would love to merge this but the build is borken :eyes:

Shredder121:
Changed the base just now, I must have forgotten that 🤔
But rebased on newest development

PR #313, opened 2017-09-27
Add support for setting the token through the CLI

Kurozume:


FredBoat/src/main/java/fredboat/FredBoat.java
```diff
@@ -101,7 +101,10 @@
 
     private static DatabaseManager dbManager;
 
+    public static String BACKUP_TOKEN;
+
     public static void main(String[] args) throws LoginException, IllegalArgumentException, InterruptedException, IOException, UnirestException {
+        BACKUP_TOKEN = args[0];
```
Shredder121: The problem is that anyone can see the command line arguments of your app. :(
I'd recommend environment variables. (`System.getEnv("fredboat.backupToken")` or something)

Frederikam:
This looks like a bad idea from a security perspective. You'd be able to see the token from htop if you specified it this way.

FredBoat/src/main/java/fredboat/Config.java
```diff
@@ -62,7 +62,7 @@
     public static int HIKARI_TIMEOUT_MILLISECONDS = 1000;
 
     private final DistributionEnum distribution;
-    private final String botToken;
+    private String botToken;
```
Shredder121: I would keep it final and extend the `else botToken = ""` into its own body for the BACKUP_TOKEN.

napstr:
1.

> Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 0
	at fredboat.FredBoat.main(FredBoat.java:107)

2.

This clashes with the legacy scope system.

Frederikam:
>This clashes with the legacy scope system.
Which we should kill

PR #315, opened 2017-09-29
Add ;;math command

RealKC:
This adds a `;;math` command which can perform the next operations: addition, subtraction, multiplication, division, modulo, square root and can calculate the power of num1 to num2 and percentages.

This depends on a Fredboat-i13n PR I need to make.

Reviews would be nice.

Also yeah I fucked accounts again, GG me.

### DIFF
--- a/FredBoat/src/main/java/fredboat/command/util/MathCommand.java
+++ b/FredBoat/src/main/java/fredboat/command/util/MathCommand.java
@@ -1,7 +1,138 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2017 Frederik Ar. Mikkelsen
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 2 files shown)


RealKC:
Yeah yeah fuck off bot

RealKC:
Reviews are welcome by the way.

napstr:
Please fix the compile error.

image

RealKC:
@napstr can check it out rn, build should fail because of the FredBoat-I13n PR

FredBoat/src/main/java/fredboat/command/util/MathCommand.java
```diff
@@ -0,0 +1,153 @@
...
+                        return;
+                }
+
+            } else {
+                HelpCommand.sendFormattedCommandHelp(context);
+                return;
+            }
+
+        } catch(NumberFormatException ex) {
+            HelpCommand.sendFormattedCommandHelp(context);
```
napstr: Should think about a more specific message here. "could not parse at least one of you numbers" or somethign like that.

RealKC:
Oh I forgot to add the #, lemme add them.

Personally not sure how to remove the trailing 0s in numbers like that.

FredBoat/src/main/java/fredboat/command/util/MathCommand.java
RealKC: Can't think of a better message, so I'll use that(unless someone has better suggestions).

Shredder121:
trimTrailingZeros?
It by default takes the same significant digits as the input arguments

PR #326, opened 2017-10-01
Added simple way to disable commands globally

Nanabell:


Nanabell:
oops pressed the wrong button

Nanabell:
😌 fixed

Issue #219, opened 2017-07-08
Unbox <links> for ;;play


napstr:
When using play like this: `;;play <https://www.youtube.com/watch?v=whatever>` (the <> boxing tells discord to not embed a link) a search happens.
Changing this to directly playing the link is probably better UX.

vince21298:
Hi. I am new to the open source community and will be my first contribution. I want to take a crack at the problem. Thank you.

napstr:
Sure thing. If you have questions about settings things up for development or want to discuss code stuff, make sure to join our discord https://discord.gg/ZWNr7G8 and poke an admin for access to the `#coding` channel.

PR #329, opened 2017-10-01
Unbox <links> for ;;play

natanbc:
See #219

### DIFF
--- a/FredBoat/src/main/java/fredboat/command/music/control/PlayCommand.java
+++ b/FredBoat/src/main/java/fredboat/command/music/control/PlayCommand.java
@@ -93,19 +93,16 @@ public class PlayCommand extends Command implements IMusicCommand, ICommandRestr
             return;
         }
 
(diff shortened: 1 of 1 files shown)


FredBoat/src/main/java/fredboat/command/music/control/PlayCommand.java
```diff
@@ -99,11 +99,16 @@ public void onInvoke(CommandContext context) {
         }
 
         //Search youtube for videos and let the user select a video
-        if (!args[1].startsWith("http")) {
+        if (!(args[1].startsWith("http") || (args[1].startsWith("<http") && args[1].endsWith(">")))) {
```
Frederikam: Why not remove the <> from the string first, then use that instead?

napstr:
Any update on this @natanbc ?

SuhanKoh:
Is anyone taking over this issue? or should it be closed?

Shredder121:
@napstr How's this now?

PR #330, opened 2017-10-01
Add Blacklist Command

Nanabell:
This is not done by any means and I don't think this is how we imagine a blacklist for the permission system but this is just a simple starting point where I want to gather ideas on how to work on this.

I just slapped the preexisting Permission commands with the BASE level permission and updated the database to store blacklisted users
We likely need a whole new Command for this as the Strings from the I13n repo don't really fit with a blacklist and rather with a whitelist.

Essentially it works as its checking the blacklist level b4 any other db bound level but it would need some from self-add prevention so that users don't blacklist themselves by accident

### DIFF
--- a/FredBoat/src/main/java/fredboat/command/moderation/BlacklistCommand.java
+++ b/FredBoat/src/main/java/fredboat/command/moderation/BlacklistCommand.java
@@ -1,0 +1,214 @@
+package fredboat.command.moderation;
+
+import fredboat.command.util.HelpCommand;
+import fredboat.commandmeta.abs.Command;
+import fredboat.commandmeta.abs.CommandContext;
+import fredboat.commandmeta.abs.IModerationCommand;
+import fredboat.db.EntityReader;
+import fredboat.db.EntityWriter;
+import fredboat.db.entity.GuildPermissions;
+import fredboat.feature.togglz.FeatureFlags;
+import fredboat.messaging.CentralMessaging;
+import fredboat.messaging.internal.Context;
+import fredboat.perms.PermissionLevel;
+import fredboat.perms.PermsUtil;
+import fredboat.shared.constant.BotConstants;
+import fredboat.util.ArgumentUtil;
+import net.dv8tion.jda.core.EmbedBuilder;
+import net.dv8tion.jda.core.Permission;
+import net.dv8tion.jda.core.entities.*;
+import net.dv8tion.jda.core.utils.PermissionUtil;
+
+import javax.annotation.Nonnull;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BlacklistCommand extends Command implements IModerationCommand {
+    @Override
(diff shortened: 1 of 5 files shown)


Nanabell:
Before I continue to work on this I will wait for feedback for #331 as this pr will directly affect this

napstr:
Calling the blacklist level BASE sounds confusing. Imo rename or add a blacklist level below it.

Nanabell:
alright fair enough, I have worked on this since but its mostly unfinished so I'm gonna push once I have something more solid with a new level

Nanabell:
Added a `BLACKLISTED` level with an int level of -1, currently its basically the same as `BASE` i dunno if we want to change this that Blacklisted users cant use the bot at all

FredBoat/src/main/java/fredboat/command/moderation/BlacklistCommand.java
```diff
@@ -0,0 +1,214 @@
...
+import net.dv8tion.jda.core.EmbedBuilder;
+import net.dv8tion.jda.core.Permission;
+import net.dv8tion.jda.core.entities.*;
+import net.dv8tion.jda.core.utils.PermissionUtil;
+
+import javax.annotation.Nonnull;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BlacklistCommand extends Command implements IModerationCommand {
```
napstr: This class looks like a lot of copy pasta of the PermissionsCommand. There is certainly a better way? At least put the copied static methods into one place.

napstr:
Some general thoughts on this:
What I like in the current design decision is it being separate from the autoblacklisting feature, as this has different goals (giving guilds a tool to deal with trolls).
Can you try making this command 100% part of the FredBoat permission system?

> i dunno if we want to change this that Blacklisted users cant use the bot at all
Definitely the way to go imo. Blacklisted users should not be able to execute any commands in the guild that blacklisted them.
However checking users for being blacklisted is a bit "tricky" cause we would have to check it on every command issued. The public FredBoat executes 10s of commands per second currently (and that's just music), so it is worth exploring optimized ways to check that value. Probably in a similar way that #375 checks prefixes without loading the whole guild config entity + caching.

KKHaneul:
when is merge this pr? T.T

napstr:
When it's done.

FredBoat/src/main/java/fredboat/command/moderation/BlacklistCommand.java
```diff
@@ -0,0 +1,214 @@
...
+            context.reply("Permissions are currently disabled.");
+            return;
+        }
+        String[] args = context.args;
+        if (args.length < 2) {
+            HelpCommand.sendFormattedCommandHelp(context);
+            return;
+        }
+
+        switch (args[1]) {
```
SuhanKoh: I think calling toLowerCase() might be a good idea here.

Issue #227, opened 2017-07-24
[BUG] Wrong error message when a role/user is already added to a specific permission level


RealKC:
![img](https://cdn.discordapp.com/attachments/242275454470324225/339150032655876098/unknown.png)

When a user/role is already added to a permission level(for example: DJ), trying to add it to DJ will say that the user/role is not found, while it should say that the user/role is already added to that permission level, something like "User/role is already `DJ`!"

Shredder121:
This also annoyed me and Ruby.
We wanted to add Ruby to a list, and I renamed myself so we would at least have a duplicate search result.
But it looks like it skips already processed users, so it added a person that happened to fulfill the search query.

RealKC:
Also something related to this is that `;;admin list` will not list the owner and roles/members that have the `Administrator` permission, despite them having the `ADMIN` permissions(since by default owners and members/roles with `Administrator` permission have all bot permissions).

Possible new title for issue: Wrong/missing info from permission management commands?

Frederikam:
The simple solution would be to say "No new roles and users" in place of "Nothing"

Frederikam:
To clarify, this is not a bug. This has to do with user experience.

napstr:
This issue is partly solved in #331.

This is what's left:

> Also something related to this is that ;;admin list will not list the owner and roles/members that have the Administrator permission, despite them having the ADMIN permissions(since by default owners and members/roles with Administrator permission have all bot permissions).

I wonder how this should be accomplished? I fear we would have to check each member of a guild for those permissions. What if there are <a lot> admins (think 1000, just an example), how are we going to show them?

PR #331, opened 2017-10-01
small fix for the Permission command

Nanabell:
This is related to #227 this improves the UX by instead of saying "Nothing found for `x`" for users that are already or are not in the list respectively 
it responds with this 
image

While I tested the feature I noticed that ArgumentUtil#getSearchTerm() still could bug out if the search term was equal to `add` or `remove`
image
So the second part of this pr completely strips the raw message from the prefix/command & the add/remove argument as well as the 2 spaces.

- [x] - This will need the merge of the I13n pr b4 this is can be merged

### DIFF
--- a/FredBoat/src/main/java/fredboat/command/moderation/PermissionsCommand.java
+++ b/FredBoat/src/main/java/fredboat/command/moderation/PermissionsCommand.java
@@ -125,7 +125,7 @@ public class PermissionsCommand extends Command implements IModerationCommand {
         if (selected == null) return;
 
         if (!gp.getFromEnum(permissionLevel).contains(mentionableToId(selected))) {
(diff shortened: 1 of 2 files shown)


napstr:
Looks good to me for a fix. Started looking at that ArgumentUtil code, but then decided to not look too close lol.
Could you switch to Context#i18nFormat() for your newly added strings please?

Nanabell:
i almost forgot about this

napstr:
Merged i13n https://github.com/Frederikam/FredBoat-i13n/pull/25. Please add here.

napstr:
Sec, after I do the crowdin rebuild.

napstr:
Done.

PR #337, opened 2017-10-02
Refactor/i18n context convenience

napstr:
You know the drill - This is a refactoring PR - don't be intimidated by the line counts, the behaviour is unchanged:tm:   :  ^)

Reasons: Check out the readability improvements in classes like SkipCommand, ConfigCommand, PermissionsCommand, VoteSkipCommand

As soon as this gets merged I'm updating contributing.md

### DIFF
--- a/FredBoat/src/main/java/fredboat/messaging/internal/Context.java
+++ b/FredBoat/src/main/java/fredboat/messaging/internal/Context.java
@@ -45,7 +45,6 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.io.File;
 import java.text.MessageFormat;
-import java.util.MissingResourceException;
 import java.util.ResourceBundle;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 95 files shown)


FredBoat/src/main/java/fredboat/messaging/internal/Context.java
```diff
@@ -138,10 +148,44 @@ public void replyPrivate(@Nonnull String message, @Nullable Consumer<Message> on
...
         return getGuild().getSelfMember().hasPermission(getTextChannel(), permissions);
     }
 
+    //return a single translated string
+    @CheckReturnValue
+    public String i18n(@Nonnull String key) {
+        try {
+            return getI18n().getString(key);
+        } catch (MissingResourceException e) { //fall back to default i18n bundle
+            return I18n.DEFAULT.getProps().getString(key);
```
Frederikam: Isn't there a better way? This is bound to have rather bad performance

napstr: Right, could make use of ResourceBundle#containsKey().
Should also log a warning for missing keys so that we can fix that.

Issue #333, opened 2017-10-02
Command to change bot avatar


Frederikam:
This would be a simple command that takes an image file and sets it as the bot's avatar.

Possible usage:
![image](https://user-images.githubusercontent.com/2582617/31069392-95b2999a-a75c-11e7-87c5-8c382bcbe71d.png)

This command would be restricted with the `BOT_ADMIN` permission.

napstr:
Better to use `;;setavatar` since there's already an `;;avatar` command

Frederikam:
Ah, my bad

yeoji:
I gave this a go and have opened a pull request :)

napstr:
#338 resolves this.

ashleykrose:
This no longer works. Was it removed?

napstr:
It was never removed.
![image](https://user-images.githubusercontent.com/6048348/43034476-6e701012-8cdd-11e8-93b8-24920b316035.png)

ashleykrose:
![capture](https://user-images.githubusercontent.com/8593303/43111017-16ef15fe-8ebd-11e8-900f-d25279296157.JPG)

ashleykrose:
it didn't set it for me

Frederikam:
1) There are no FredBoat with write access to that channel, as it would respond
2) You need to be a bot admin, which you certainly aren't of the main FredBoat

ashleykrose:
how to i do that?

ashleykrose:
![capture](https://user-images.githubusercontent.com/8593303/43111523-ad52a68a-8ebf-11e8-9d90-eeb79477e3de.JPG)

Frederikam:
Host your own bot if you want to change the avatar

PR #338, opened 2017-10-02
Add set avatar command (#333)

yeoji:
Implemented the Set Avatar command described in issue #333 
Fixes: #333 

There seems to be a limit to how many times the avatar can be changed in a set period of time (probably around 2 times every ~15 minutes).

### DIFF
--- a/FredBoat/src/main/java/fredboat/command/admin/SetAvatarCommand.java
+++ b/FredBoat/src/main/java/fredboat/command/admin/SetAvatarCommand.java
@@ -1,0 +1,74 @@
+package fredboat.command.admin;
+
+import com.mashape.unirest.http.Unirest;
+import com.mashape.unirest.http.exceptions.UnirestException;
+import fredboat.FredBoat;
+import fredboat.commandmeta.abs.Command;
+import fredboat.commandmeta.abs.CommandContext;
+import fredboat.commandmeta.abs.ICommandRestricted;
+import fredboat.messaging.internal.Context;
+import fredboat.perms.PermissionLevel;
+import net.dv8tion.jda.core.entities.Icon;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nonnull;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * This command allows a bot admin to change the avatar of FredBoat
+ */
+public class SetAvatarCommand extends Command implements ICommandRestricted {
+
+    private static final Logger log = LoggerFactory.getLogger(SetAvatarCommand.class);
+
+    @Override
+    public void onInvoke(CommandContext context) {
+        if (context.args.length > 1) {
+            String imageUrl = context.args[1];
+
+            if (imageUrl.startsWith("http://") || imageUrl.startsWith("https://")) {
+                String type = getImageMimeType(imageUrl);
+
+                if (type.equals("image/jpeg") || type.equals("image/png") || type.equals("image/gif") || type.equals("image/webp")) {
+                    setBotAvatar(imageUrl);
+                }
+            }
+        }
+    }
+
+    @Nonnull
+    @Override
+    public String help(@Nonnull Context context) {
+        return "{0}{1} <imageUrl>\n#Sets the bot avatar to the image file.";
+    }
+
+    private String getImageMimeType(String imageUrl) {
+        String mimeType = "";
+        try {
+            mimeType = Unirest.get(imageUrl).asBinary().getHeaders().getFirst("Content-Type");
+        } catch (UnirestException e) {
+            log.error("Error retrieving mime type of image!", e);
+            throw new RuntimeException(e);
+        }
+        return mimeType;
+    }
+
+    private void setBotAvatar(String imageUrl) {
+        InputStream avatarData;
+        try {
+            avatarData = Unirest.get(imageUrl).asBinary().getBody();
+            FredBoat.getFirstJDA().getSelfUser().getManager().setAvatar(Icon.from(avatarData)).queue();
+        } catch (UnirestException e) {
+            throw new RuntimeException(e);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public PermissionLevel getMinimumPerms() {
+        return PermissionLevel.BOT_ADMIN;
+    }
+}
--- a/FredBoat/src/main/java/fredboat/command/admin/SetAvatarCommand.java
+++ b/FredBoat/src/main/java/fredboat/command/admin/SetAvatarCommand.java
@@ -36,18 +36,13 @@ public class SetAvatarCommand extends Command implements ICommandRestricted {
             imageUrl = context.args[1];
         }
 
-        if (imageUrl == null) {
-            HelpCommand.sendFormattedCommandHelp(context);
-        } else {
-            if (imageUrl.startsWith("http://") || imageUrl.startsWith("https://")) {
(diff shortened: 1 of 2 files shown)


napstr:
The JDA discord lib which we are using provides methods to do this in a clean way. Please use those instead instead of hand writing API requests.

yeoji:
I've updated the code to use the JDA lib

FredBoat/src/main/java/fredboat/command/admin/SetAvatarCommand.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Override
+    public void onInvoke(CommandContext context) {
+        if (context.args.length > 1) {
+            String imageUrl = context.args[1];
+
+            if (imageUrl.startsWith("http://") || imageUrl.startsWith("https://")) {
+                String type = getImageMimeType(imageUrl);
+
+                if (type.equals("image/jpeg") || type.equals("image/png") || type.equals("image/gif")) {
```
Frederikam: Please add `image/webp` which is what Discord typically uses

napstr:
There was a change to one of the command methods signature - the help method receives a Context instead of a guild now. Could you please adjust your PR for that - sorry for the inconvenience, I tried merging all relevant and active open PRs before that to avoid breaking things but yours popped up surprisingly today :)

yeoji:
Added "image/webp" as an accepted type and also rebased onto the development branch to align this PR with the new changes :)

yeoji:
Success and failure messages have been added and the code has been realigned with development branch to resolve conflicts.

FredBoat/src/main/java/fredboat/command/admin/SetAvatarCommand.java
```diff
@@ -0,0 +1,82 @@
...
+import java.io.InputStream;
+
+/**
+ * This command allows a bot admin to change the avatar of FredBoat
+ */
+public class SetAvatarCommand extends Command implements ICommandRestricted {
+
+    private static final Logger log = LoggerFactory.getLogger(SetAvatarCommand.class);
+
+    private CommandContext context;
```
napstr: Why does this field exist? The CommandContext provides values that are interesting during command execution only. It also holds direct references to JDA objects, which should not be held as long term references since they tend to become outdated and cause memory leaks if shards are revived. If you really need to save a context for later use (when you are writing a more complex feature), you will need to use a LeakSafeContext / one of it's existing child classes.

yeoji: My bad, I've changed this to pass the context into the method that uses it.

Shredder121:
@napstr Are we (potentially) merging this before #343?

napstr:
Yes, this should be merged before #343

FredBoat/src/main/java/fredboat/command/admin/SetAvatarCommand.java
```diff
@@ -0,0 +1,78 @@
...
+            String imageUrl = context.args[1];
+
+            if (imageUrl.startsWith("http://") || imageUrl.startsWith("https://")) {
+                String type = getImageMimeType(imageUrl);
+
+                if (type.equals("image/jpeg") || type.equals("image/png") || type.equals("image/gif") || type.equals("image/webp")) {
+                    setBotAvatar(context, imageUrl);
+                }
+            }
+        }
```
napstr: This silently fails when there are not enough arguments provided (0) or the regex doesn't pick up a link. What should happen instead is that the command help should be posted if users use a command wrong and we notice that.

This is also only solves half of the original issue, as it only accepts links, but not uploaded files.

FredBoat/src/main/java/fredboat/command/admin/SetAvatarCommand.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Nonnull
+    @Override
+    public String help(@Nonnull Context context) {
+        return "{0}{1} <imageUrl>\n#Sets the bot avatar to the image file.";
```
napstr: This is not correct. The current implementation only uses links.
This is how the full text should be with the full implementation:
`{0}{1} <imageUrl> OR <file>\n#Set the bot avatar to the image provided by url or file.`

Frederikam: Maybe "file" should be "attachment" for the sake of clarification

yeoji: How does using an attachment in the command work? I tried attaching an image file, and then Copied the link and used that in the setavatar command and it's still a normal http URL.

screen shot 2017-10-05 at 11 08 16 pm

FredBoat/src/main/java/fredboat/command/admin/SetAvatarCommand.java
yeoji: My bad, I will add the Help message now. Also please see my comment below as I'm not sure how the command should accept uploaded files.

FredBoat/src/main/java/fredboat/command/admin/SetAvatarCommand.java
Frederikam: The first image is an attachment, the second is just a link. IIRC you need the Message#getAttachments() endpoint

yeoji: Hmm, can you please give me an example of how the user would use the command? Eg. <<setavatar some-avatar.png?

Frederikam: I tried explaining that in the original issue, but it's like this
image

yeoji: Ohh, got it! Thanks :)

screen shot 2017-10-05 at 11 26 54 pm

FredBoat/src/main/java/fredboat/command/admin/SetAvatarCommand.java
```diff
@@ -0,0 +1,91 @@
...
+
+        if (imageUrl == null) {
+            HelpCommand.sendFormattedCommandHelp(context);
+        } else {
+            if (imageUrl.startsWith("http://") || imageUrl.startsWith("https://")) {
+                String type = getImageMimeType(imageUrl);
+
+                if (type.equals("image/jpeg") || type.equals("image/png") || type.equals("image/gif") || type.equals("image/webp")) {
+                    setBotAvatar(context, imageUrl);
+                }
```
napstr: This whole if construct is still having several branches where it silently fails.

You probably should invert the flow, and return if setBotAvatar() is successful, otherwise have it fall through to send the help.

FredBoat/src/main/java/fredboat/command/admin/SetAvatarCommand.java
```diff
@@ -0,0 +1,91 @@
...
+        try {
+            avatarData = Unirest.get(imageUrl).asBinary().getBody();
+            FredBoat.getFirstJDA().getSelfUser().getManager().setAvatar(Icon.from(avatarData))
+                    .queue(
+                            success -> context.reply("Avatar has been set successfully!"),
+                            failure -> context.reply("Error setting avatar. Please try again later.")
+                    );
+        } catch (UnirestException e) {
+            throw new RuntimeException(e);
+        } catch (IOException e) {
```
napstr: Here's a neat way to group exceptions if you are going to handle them the same way anyways:
```java
        } catch (UnirestException | IOException e) {
            throw new RuntimeException(e);
        }
```

PR #339, opened 2017-10-02
Added HttpSourceManager as replacement for HttpAudioSourceManager.

### LINKED ISSUE #43: .m3u and .pls web scraper
FredBoat currently supports live internet radios, but not that many people are using it. It seems that many are not aware how to connect FredBoat to a radio. Current usage is like this:

```
;;play https://gensokyoradio.net/GensokyoRadio.pls
;;play https://gensokyoradio.net/GensokyoRadio.m3u
;;play http://listen.moe/stream.m3u
[etc]
```

Lavaplayer needs a direct link to the playlist file and will fail to load such URLs like `https://gensokyoradio.net` since it returns an HTML page.

I propose the idea that we regex any failed URL identifier for m3u and pls links after lavaplayer has failed to load the identifier as a URL in the `AudioLoader`. The following regex could be used:
```js
/href=\"([\w\d\.\/]{1,}?\.(?:m3u|pls))\"/g
```

kantenkugel:
It first tries to get audio source directly from given link like before.
If that fails and http connection has content-type text/html, it searches the html body for playlist links 
(m3u and pls) and fetches audio-source from them

Closes #43

### DIFF
--- a/ThirdPartyLicenses/APACHE2
+++ b/ThirdPartyLicenses/APACHE2
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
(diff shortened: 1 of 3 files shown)


kantenkugel:
Fancy CI check

ThirdPartyLicenses/APACHE2
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "{}"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright {yyyy} {name of copyright owner}
```
Frederikam: This doesn't look right

kantenkugel: its a 1:1 copy of his license file. @sedmelluq doesn't have licensing headers in the files either but i asked him if that attribution in the file is ok for him

kantenkugel:
Oh... i throw some exceptions in my code. the Common one will get printed to the user (via handleThrowable of AudioLoader). Should i localize that exception message?

Frederikam:
Where?

If we fail to find anything or to connect, we should not load any tracks
image

kantenkugel:
Not sure what you meant... output in image not ok?

Frederikam:
>No audio could be found for https://github.com/Frederikam/FredBoat/pull/339.

This means no AudioTrack was loaded by any of the source managers, which is the intended behaviour for a bad HTTP link

kantenkugel:
There are 3 exceptions that i throw:
- Could not read inputstream into string
- Could not create URL object from identifier even tho it is a valid url
- Could not resolve the embedded link to playlist against the current url (link in href="" doesn't start with http, .(dot) or /)

Only the third one's message will be printed (currently unlocalized "URL could not be resolved")
The 2nd and 3rd one also only ever happen if there is actually a html site with some playlist link embedded.

Frederikam:
Would you consider these to be rare exceptions?

kantenkugel:
not so sure about 3rd one, but first 2 would be pretty rare i think (http stream is open already and was successfully probed, it is obviously a url (the html response comes from there)

kantenkugel:
i can make the 3rd one be fallthrough (nothing found)

Frederikam:
Please do

kantenkugel:
explained a bit more in previous post (edited)

Frederikam:
Alright fair enough. Leave it as it is, without I18n

kantenkugel:
Ok im going to allow it to handle paths that start with chars as well (`href="relPath.m3u"`) and make default fall-through if it doesn't match dot, / , http or starting with char

PR #340, opened 2017-10-02
Fix dance command having more than one running instance

Shredder121:
This uses concurrent locks, to lock across threads, instead of object monitors, which belong to a thread, and can't be queried whether they are held or not.

### DIFF
--- a/FredBoat/src/main/java/fredboat/command/fun/DanceCommand.java
+++ b/FredBoat/src/main/java/fredboat/command/fun/DanceCommand.java
@@ -25,41 +25,58 @@
 
 package fredboat.command.fun;
 
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
 import fredboat.commandmeta.abs.Command;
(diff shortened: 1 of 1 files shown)


FredBoat/src/main/java/fredboat/command/fun/DanceCommand.java
```diff
@@ -25,41 +25,58 @@
...
 import javax.annotation.Nonnull;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Function;
 
 public class DanceCommand extends Command implements IFunCommand {
 
+    private final Function<Guild, ReentrantLock> locks = CacheBuilder.newBuilder()
+            .maximumSize(128) //any value will do, but not too big
```
Shredder121: For posterity, any value will do, since the buckets (`maximumSize/concurrencyLevel`) will maintain the limit separately.
This number was just chosen so that 1) it won't run to infinite entries and 2) enough people will be able to use the dance command.

PR #290, opened 2017-09-11
Add ;;weather command

SuhanKoh:
**Sample output on the text channel**

``;;weather san francsico``
weather1

``;;weather singapore``
weather2

PR #343, opened 2017-10-03
⚰ Unirest

napstr:
Unirest is an unmaintained http lib. I didn't waste much thought into choosing a replacement, just picked okhttp as JDA uses it too. The thing about okhttp is while Unirest is static hell, OkHttp swings out really strong in the opposite direction being extremely object oriented which can be quite annoying at times, so part of this PR is a unirest-like okhttp wrapper [Http.java](https://github.com/Frederikam/FredBoat/commit/27eb8739d12af18b25f33fc278b67d89e73e7635#diff-2de669976a7936d93331ea8d5d69811c) which is good enough:tm: for our most common calls.

Looking for general feedback at this point.
Do not merge this before #290 and #338

I did test each of the replaced calls individually and all of them appear to be working, except the stats post to Carbonitex which I can't test since I do not own a Carbonitex key.

### DIFF
--- a/FredBoat/src/main/java/fredboat/agent/CarbonitexAgent.java
+++ b/FredBoat/src/main/java/fredboat/agent/CarbonitexAgent.java
@@ -29,6 +29,7 @@ import fredboat.Config;
 import fredboat.FredBoat;
 import fredboat.util.rest.Http;
 import net.dv8tion.jda.core.JDA;
+import okhttp3.Response;
 import org.slf4j.LoggerFactory;
 
 import java.util.concurrent.TimeUnit;
@@ -65,15 +66,20 @@ public class CarbonitexAgent extends FredBoatAgent {
             return;
         }
 
-        try {
-            final String response = Http.post("https://www.carbonitex.net/discord/data/botdata.php",
+        try (Response response = Http.post("https://www.carbonitex.net/discord/data/botdata.php",
                     Http.Params.of(
                             "key", key,
                             "servercount", Integer.toString(FredBoat.countAllGuilds())
                     ))
-                    .asString();
+                .execute()) {
 
-            log.info("Successfully posted the bot data to carbonitex.com: " + response);
+            //noinspection ConstantConditions
+            String content = response.body().string();
+            if (response.isSuccessful()) {
+                log.info("Successfully posted the bot data to carbonitex.com: {}", content);
+            } else {
+                log.warn("Failed to post stats to Carbonitex: {}\n{}", response.toString(), content);
+            }
         } catch (Exception e) {
             log.error("An error occurred while posting the bot data to carbonitex.com", e);
         }
--- a/FredBoat/src/main/java/fredboat/agent/CarbonitexAgent.java
+++ b/FredBoat/src/main/java/fredboat/agent/CarbonitexAgent.java
@@ -25,9 +25,9 @@
 
 package fredboat.agent;
 
-import com.mashape.unirest.http.Unirest;
 import fredboat.Config;
 import fredboat.FredBoat;
+import fredboat.util.rest.Http;
 import net.dv8tion.jda.core.JDA;
 import org.slf4j.LoggerFactory;
 
@@ -66,10 +66,13 @@ public class CarbonitexAgent extends FredBoatAgent {
         }
 
         try {
-            final String response = Unirest.post("https://www.carbonitex.net/discord/data/botdata.php")
-                    .field("key", key)
-                    .field("servercount", FredBoat.countAllGuilds())
-                    .asString().getBody();
+            final String response = Http.post("https://www.carbonitex.net/discord/data/botdata.php",
+                    Http.Params.of(
+                            "key", key,
+                            "servercount", Integer.toString(FredBoat.countAllGuilds())
+                    ))
+                    .asString();
+
             log.info("Successfully posted the bot data to carbonitex.com: " + response);
         } catch (Exception e) {
             log.error("An error occurred while posting the bot data to carbonitex.com", e);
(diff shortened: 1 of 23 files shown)


FredBoat/src/main/java/fredboat/agent/CarbonitexAgent.java
```diff
@@ -66,10 +66,13 @@ private void sendStats() {
...
-            final String response = Unirest.post("https://www.carbonitex.net/discord/data/botdata.php")
-                    .field("key", key)
-                    .field("servercount", FredBoat.countAllGuilds())
-                    .asString().getBody();
+            final String response = Http.post("https://www.carbonitex.net/discord/data/botdata.php",
+                    Http.Params.of(
+                            "key", key,
+                            "servercount", Integer.toString(FredBoat.countAllGuilds())
+                    ))
+                    .asString();
```
Shredder121: Do we not need the `Response` and checking whether it was successful? 🤔

FredBoat/src/main/java/fredboat/command/fun/RandomImageCommand.java
```diff
@@ -64,20 +69,15 @@ public RandomImageCommand(String[] urls) {
...
+    public RandomImageCommand(@Nonnull String imgurAlbumUrl) {
+        //update the album every hour
+        imgurRefresher.scheduleAtFixedRate(() -> {
+            try {
+                populateItems(imgurAlbumUrl);
+            } catch (Exception e) {
+                log.error("Populating imgur album {} failed", imgurAlbumUrl, e);
             }
-        }, RandomImageCommand.class.getSimpleName() + " imgur updater").start();
+        }, 0, 1, TimeUnit.HOURS);
```
Shredder121: Doing a bit much in one commit 👀

FredBoat/src/main/java/fredboat/command/util/MALCommand.java
```diff
@@ -64,38 +73,43 @@ public void onInvoke(@Nonnull CommandContext context) {
...
+            // we will try a user search instead
         }
+
+        request = Http.get("http://myanimelist.net/search/prefix.json",
+                Http.Params.of(
+                        "type", "user",
+                        "keyword", term
+                ))
+                .auth(Credentials.basic(Config.CONFIG.getMalUser(), Config.CONFIG.getMalPassword()))
+                .client(malHttpClient);
```
Shredder121: basically at this point I see duplicate code, with setting the auth and client again.
So have you tested that this works?
(I'd say, maybe have the code work a little bit more like RestTemplate, where you have all the necessary bits in a HttpRequester, instead of static methods that you call on Http and such)

FredBoat/src/main/java/fredboat/util/rest/Http.java
```diff
@@ -0,0 +1,195 @@
...
+        //pass pairs of strings and you'll be fine
+        @Nonnull
+        @CheckReturnValue
+        public static Params of(@Nonnull String... pairs) {
+            if (pairs.length % 2 == 1) {
+                log.warn("Passed an uneven number of args to the Params wrapper, this is a likely bug.");
+            }
+            Params result = new Params();
+            for (int i = 0; i <= pairs.length / 2; i += 2) {
+                result.params.put(pairs[i], pairs[i + 1]);
```
Shredder121: Apache Commons Collections has a MapUtils.putAll function.
If we don't have that dependency, you can also inline that, so you have less math going on in your brain trying to decide if it's safe.
```java
for (int i = 0; i < array.length - 1;) {
    // these casts can fail if array has incorrect types
    map.put((K) array[i++], (V) array[i++]);
}
```

FredBoat/src/main/java/fredboat/command/fun/RandomImageCommand.java
napstr: Oh no...how did this slip in :eyes:

FredBoat/src/main/java/fredboat/agent/CarbonitexAgent.java
napstr: Good point, I'll change it to do a check.

FredBoat/src/main/java/fredboat/command/util/MALCommand.java
napstr: Yeah this method looks like shit due to the MAL API being really bad and timing out on some requests.

I suppose we could have a method on the SimpleRequest that sets the request url.
I don't really know the RestTemplate stuff, do you mind sharing a code example?

napstr: > So have you tested that this works?

Here's the weird thing: If we do the anime search first, and it returns no result, then the subsequent user search will time out.
I can even reproduce this is a REST tool like Insomnia. Firing user searches works fine. But don't you dare search for an anime, the user searches will start timing out.

wat

napstr: Ok, not time out. Got a result back after 1 minute. Some kind of weird rate limit?

napstr: Ok more results from testing: so if you do an anime search that doesn't return a result, all your following requests will only be answered after 1 minute until you do a successful request. <:SmilingBlackGuyWithQuestionMarks:274075979377410048>


This is some serious fuckery and I'm not finding any resources on this topic to fix it straight away. Marking the MAL command as deprecated and opening an issue.

napstr:
> Maybe have a built instance per place that needs to send requests.

Using those where needed. For example as part of this PR the MAL command got a special one with different timeouts.
That is actually one of the reasons Unirest needs to go.

PR #346, opened 2017-10-04
Use a single queue for both connecting and reconnecting Fixes #303

### LINKED ISSUE #303: Hook coin system up to JDAs reconnect queue
Hitting ratelimits during start up when shards have to reconnect. Hook up coin system / shard creation to JDAs reconnect queue to solve this.

https://sentry.davinci.sh/fredboat/fredboat/issues/2454/

```
Encountered IDENTIFY (OP 2) Rate Limit! Waiting 5 seconds before trying again!
```

napstr:
This needs opinions. It looks ugly but it works.

The main issue I see here is that in case JDA changes the implementation of their ReconnectQueue we might be forced to adapt ours to update, however I didn't manage to find a more independent solution with the way the ReconnectQueue is written (running its own workers, calling protected methods).

Fixes #303

### DIFF
--- a/FredBoat/src/main/java/fredboat/FredBoat.java
+++ b/FredBoat/src/main/java/fredboat/FredBoat.java
@@ -73,7 +73,6 @@ public abstract class FredBoat {
 
     private static final Logger log = LoggerFactory.getLogger(FredBoat.class);
 
-    public static final int SHARD_CREATION_SLEEP_INTERVAL = 5500;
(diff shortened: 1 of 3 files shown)


napstr:
This is done. Tested locally by starting with 99 shards and pulling the network a few times.

napstr:
Deployed this during the network issues earlier and it worked fine.

PR #347, opened 2017-10-04
Reenable full commands list for selfhosters

napstr:
This is no mistake I want this to be merged into stable.
Pl0x merge stable into dev afterwards.


napstr:
Commited this directly through github a8258f2e8ec45958db44075687972c31893e1e46. Apparently the master branch is only locked from remote commits.

Shredder121:
If it's only branch protection, it's only protecting against force pushes, accidental or otherwise.
If statuses/reviews were required, you'd not be able to.

Issue #114, opened 2017-03-21
[suggestion] Play selects multiple songs at once 


Yeetoo:
after using 
_;;play (search)_
we would use 
_;;play (choices separated by commas)_

Frederikam:
Eg like *;;play 1,2, 5* would select songs 1, 2 and 5 in a selection

SuhanKoh:
I can work on this if no one has started on this.

PR #349, opened 2017-10-05
Implemented multi-select option

SuhanKoh:
This is for #114


SuhanKoh:
Question: Not sure if we should change the i18n for 'selectSuccess', currently when you select multiple songs, it will print the first one.

napstr:
Hi, I would appreciate it if you could please not reformat whitespace unrelated to the lines you are changing in your PRs as it's a possible source of unnecessary conflicts.

napstr:
If you are lazy like me you can achieve that by using the IntelliJ SaveActions plugin with a ticked "reformat only changed code" option.
image or manually staging hunks when creating the commit and discarding those lines that the reformatter was too eager to touch.

napstr:
This isn't in a working state at all. Here are a few broken things, this is most likely not a full list, as I didn't event try breaking it on purpose.

Detection not working with the numbered alias:
image

Using the number alias selects only 1 instead of several:
image

Unhandled exceptions:
image

No response to the user after selecting:
image


Please test your code more extensively.

SuhanKoh:
Fix the following

- [ ] <<1,2,3,4,5 

- [ ] <<1 2 3 4 5 should have included all 5 searches

- [ ] <<p 1 2 3 should not throw error

- [ ] <<p 1,2,3,4,5 should not have duplicated 1s.

SuhanKoh:
@napstr So ;;p 1 2 3 4 5 will queue all 5 songs? based on #114 it specified comma

napstr:
> @napstr So ;;p 1 2 3 4 5 will queue all 5 songs? based on #114 it specified comma

That's just one example.
Personally I would suggest going as far as allowing 12345 or 145 or 314 or whatever combinations.

SuhanKoh:
Should the command notify if one of the selection is dropped? eg: ``<<1 23 2 4`` 23 would be ignored?

napstr:
I think telling the user which tracks were selected is an easy and transparent way to tell them what's happening.

Example:
Input:
```
;;2 1 5
```
Output:
```
Song #2 has been selected: Some song (02:10)
Song #1 has been selected: Whatever (04:52)
Song #5 has been selected: Another song (13:28)
```

SuhanKoh:
Should I change the output to Embed then? because otherwise multi-select won't look good as the first song get printed on first line

SuhanKoh:
Print out example:
capture

PR #352, opened 2017-10-06
Rework/jda entity counts

napstr:
I would be real happy to get a review from @Shredder121 on this one since you worked last time on this.

### DIFF
--- a/FredBoat/src/main/java/fredboat/FredBoat.java
+++ b/FredBoat/src/main/java/fredboat/FredBoat.java
@@ -64,7 +64,6 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.security.auth.login.LoginException;
 import java.io.IOException;
-import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutorService;
@@ -86,8 +85,9 @@ public abstract class FredBoat {
     //central event listener that all events by all shards pass through
     protected static EventListenerBoat mainEventListener;
     protected static final ConnectQueue connectQueue = new ConnectQueue();
-    protected final static StatsAgent uniqueUsersCounterAgent = new StatsAgent("unique users counter");
+    protected final static StatsAgent jdaEntityCountAgent = new StatsAgent("jda entity counter");
 
+    private final static JdaEntityCounts jdaEntityCountsTotal = new JdaEntityCounts();
     private static DatabaseManager dbManager;
     private static final List<FredBoat> shards = new CopyOnWriteArrayList<>();
 
@@ -161,7 +161,7 @@ public abstract class FredBoat {
         //Check OpenWeather key
         executor.submit(FredBoat::hasValidOpenWeatherKey);
 
-        FredBoatAgent.start(uniqueUsersCounterAgent);
+        FredBoatAgent.start(jdaEntityCountAgent);
 
         /* Init JDA */
         initBotShards(mainEventListener);
@@ -171,7 +171,10 @@ public abstract class FredBoat {
             FredBoatAgent.start(new CarbonitexAgent(carbonKey));
         }
 
-        uniqueUsersCounterAgent.addAction(FredBoat::countUniqueUsersTotal);
+        //attempt to do counts for all shards
+        // the last ones might not be ready yet, in which case the stats agent will take care of it later
+        jdaEntityCountsTotal.count(shards);
+        jdaEntityCountAgent.addAction(() -> jdaEntityCountsTotal.count(shards));
     }
 
     // ################################################################################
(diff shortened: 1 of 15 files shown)


Shredder121:
Ran this through a profiler, noticed the same spike allocation issue that we had to fix before.
<details>
<summary>Runtimes are about the same</summary>

image
</details>

<details>
<summary>Allocations are much worse though, and this is with only `43384` users with my bot</summary>

image
</details>

napstr:
Allocations look fine to me.
This is how it looks when run together each time:
image

It does look worse in CPU time tho
image

FredBoat/src/main/java/fredboat/FredBoat.java
```diff
@@ -338,62 +339,37 @@ public JDA getJda() {
...
-    public static int getExpectedUserCount() {
-        if (biggestUserCount.get() <= 0) { //initialize
-            countAllUniqueUsers();
-        }
-        return biggestUserCount.get();
-    }
-
-    public static long countAllUniqueUsers() {
-        return JDAUtil.countAllUniqueUsers(shards, biggestUserCount);
+    private static AtomicInteger biggestUserCountTotal = new AtomicInteger(-1); //of FredBoat total
```
Shredder121: Can be final 🤔

FredBoat/src/main/java/fredboat/FredBoat.java
```diff
@@ -439,8 +414,12 @@ public long getGuildCount() {
         return JDAUtil.countAllGuilds(Collections.singletonList(this));
     }
 
+    private AtomicInteger biggestUserCount = new AtomicInteger(-1); //of this shard
```
Shredder121: Can be final 🤔

napstr:
> Are we also going to do something so it doesn't count them all the time, but only once every x time?

Do you think this is a good case to write an agent, that traverses the shards each X time, and does the calculations? Where should this data be kept, fields of the shards, or some other central place?

napstr:
Added an agent that counts unique users each minute.

napstr:
Agent counts a range of JDA entities now. Counts get updated when all shards that the count is requested for are connected to avoid wrong data points.

Shredder121:
will have to retry running it, potentially tomorrow or over the weekend

FredBoat/src/main/java/fredboat/agent/StatsAgent.java
```diff
@@ -0,0 +1,74 @@
...
+        countActions.add(action);
+    }
+
+    @Override
+    protected void doRun() {
+        for (Action action : countActions) {
+            try {
+                action.act();
+            } catch (Exception e) {
+                log.error("Unexpected exception when counting", e);
```
Shredder121: I feel like this is not a great logging statement.
Since all the `Action`s are implemented as lambdas, so will not make much sense in logs.
```java
log.error("Unexpected exception when counting {}", action.getName(), e);
//...
@FunctionalInterface
public interface Action {
    default String getName() {
        return "stats";
    }
    void act() throws Exception;
}
```
Have `stats` as default, but override them, maybe like this?
```java
//inside FredBoatShard.java
private class ShardStatsCounter implements StatsAgent.Action {
    private final JDA.ShardInfo shardInfo;
    private final StatsAgent.Action action;
    ShardStatsCounter(JDA.ShardInfo shardInfo, StatsAgent.Action action) {
        this.shardInfo = shardInfo;
        this.action = action;
    }
    @Override
    public String getName() {
        return "stats for shard" + getJda().getShardInfo();
    }
    @Override
    public void act() throws Exception {
        action.act();
    }
}
```

And have something comparable for the global stats.

Anyway, what do you think?

napstr: That looks like a decent improvement, I'll see that I add it.

napstr: Not sure if keeping a JDA shard info around might be a source of leaks when using revives? Since it's not a static class of the JDA object.

Shredder121: JDA is an interface, so every member (class) is static already.

FredBoat/src/main/java/fredboat/FredBoat.java
```diff
@@ -157,13 +162,21 @@ public static void main(String[] args) throws LoginException, IllegalArgumentExc
...
         initBotShards(mainEventListener);
 
         String carbonKey = Config.CONFIG.getCarbonKey();
         if (Config.CONFIG.getDistribution() == DistributionEnum.MUSIC && carbonKey != null && !carbonKey.isEmpty()) {
             FredBoatAgent.start(new CarbonitexAgent(carbonKey));
         }
+
+        //attempt to do counts for all shards
+        // the last ones might not be ready yet, in which case the stats agent will take care of it later
+        jdaEntityCountsTotal.count(shards);
```
Shredder121: Should this also be a `, true)`?

napstr: Nope. That force flag is only necessary cause JDA calls onReady() before setting it's status to connected, but it has technically loaded all entities.

At this place we don't know whether all shards are loaded yet. The only way would be to loop through them and check their status.
Which is actually a thing I was planning to add in the Metrics PR as it gets actually important there (to have less ugly graphs cause of inconsistent scrapes during start up). Might as well add it here.

PR #363, opened 2017-10-14
Always use the track uri in the export

Shredder121:
This is actually the same as the custom identifier in the YouTube case but makes all tracks exportable.


Shredder121:
Nice! Thanks.

PR #366, opened 2017-10-17
Limit the dance command to a number of instances

Shredder121:
Rather naive way of limiting the number of dance commands, but had this one still for some time.

### DIFF
--- a/FredBoat/src/main/java/fredboat/command/fun/DanceCommand.java
+++ b/FredBoat/src/main/java/fredboat/command/fun/DanceCommand.java
@@ -60,7 +60,9 @@ public class DanceCommand extends Command implements IFunCommand {
         // any one lock is only set-up by one thread we can get away with a naive isLocked check
         ReentrantLock lock = locks.apply(context.getGuild());
         if (lock.isLocked() || !allowed.tryAcquire()) {
-            return; //already in progress or not allowed
+            //already in progress or not allowed
+            context.reply(context.i18n("tryLater"));
+            return;
         }
         Runnable func = new Runnable() {
             @Override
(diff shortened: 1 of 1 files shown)


FredBoat/src/main/java/fredboat/command/fun/DanceCommand.java
```diff
@@ -49,15 +50,17 @@
...
     @Override
     public void onInvoke(@Nonnull CommandContext context) {
         //locking by use of java.util.concurrent.locks
 
         //in most cases we would need to use a fair lock, but since
         // any one lock is only set-up by one thread we can get away with a naive isLocked check
         ReentrantLock lock = locks.apply(context.getGuild());
-        if (lock.isLocked()) {
-            return; //already in progress
+        if (lock.isLocked() || !allowed.tryAcquire()) {
```
napstr: Wouldn't this pass in case of lock.isLocked() == false without acquiring the semaphore?

Shredder121: LOL yeah that should not happen, thanks for reviewing >.<

Shredder121: Well, actually it should be fine, if either condition is true it shouldn't dance.
What do you mean by pass?

napstr: It will pass/advance this check without bumping the semaphore, if there is no guild lock.

napstr: Oh wait...this looks fine. The || operator short circuits when the left side is true. My bad.

FredBoat/src/main/java/fredboat/command/fun/DanceCommand.java
```diff
@@ -49,15 +50,17 @@
...
+            return; //already in progress or not allowed
```
napstr: Let's instead of silenty nope'ing use this:
`context.reply(context.i18n("tryLater"));`

Issue #92, opened 2017-03-12
Handle markdown in externally generated content


napstr:
Some things that FredBoat displays to users can contain markdown syntax and fuck around with how the bots output looks:

Example:

![image](https://cloud.githubusercontent.com/assets/6048348/23833842/0abb111a-074c-11e7-8b89-adee9ce7b6ad.png)

Song names, playlist names, user names. Anything I'm missing?
Need to either strip external content of markdown tags, or maybe find an acceptable way to display it in `code` tags.

Frederikam:
This is something I have been meaning to do for quite a while. It can be solved by escaping markdown symbols with `\`

Nanabell:
To bring this up again, since we talked about preventing pings like @everyone or @here or just userpings. this is basically the same case here. Escaping @ # 
But iirc we would need to replace all sendMessage's

BadJukeBox:
Hey! I'd like to take a go at this one if you don't mind. That being said I'm a bit curious as to what the issue is, are you referring to the way that the lines are different fonts  or is that intended. Sorry if i'm missing something here. As well thank you for putting a 'how to contribute' post for us new guys :)

napstr:
[Discord supports markdown for formatting posts.](https://support.discordapp.com/hc/en-us/articles/210298617-Markdown-Text-101-Chat-Formatting-Bold-Italic-Underline-), and we make use of that to format our output nicely, example:
![image](https://user-images.githubusercontent.com/6048348/31053403-9f524c54-a69c-11e7-9885-e5bc073f263b.png)

Notice the songname and username and a few other information being bold or in code tags for better readability. However, we are using user generated content (the nickname for example), and that might contain formatting characters. Watch what happens to the same list of songs if I change my nickname:
![image](https://user-images.githubusercontent.com/6048348/31053418-04879dea-a69d-11e7-98fc-ef1b0065e63a.png)

Frederikam:
To elaborate, it is possible to escape markdown like so:


\\\_\\\*\\\*Napster\\\_

Which generates:

\_\*\*Napster\_

So simply preface any \*\_\~ with \\

jnkdev:
@BadJukeBox how's your progress on this? I'd like to try solve this too if you don't mind.

BadJukeBox:
@jnkdev Unfortunately, midterm season is in full swing for me so I haven't been able to anything that I actually want to. You can take a crack at it if you want, I play to try sometime this week myself!

[...]

PR #367, opened 2017-10-19
fixed issue #92 for markdown characters

BadJukeBox:
Added method removeMarkdown to TextUtils

username is now passed through this method before being used in ListCommand


FredBoat/src/main/java/fredboat/util/TextUtils.java
```diff
@@ -195,11 +196,11 @@ public static long parseTimeString(String str) throws NumberFormatException {
         m.find();
 
         int capturedGroups = 0;
-        if(m.group(1) != null) capturedGroups++;
-        if(m.group(2) != null) capturedGroups++;
-        if(m.group(3) != null) capturedGroups++;
+        if (m.group(1) != null) capturedGroups++;
+        if (m.group(2) != null) capturedGroups++;
+        if (m.group(3) != null) capturedGroups++;
```
napstr: Please don't format lines of code that you weren't doing any changes to.

FredBoat/src/main/java/fredboat/util/TextUtils.java
```diff
@@ -227,6 +228,25 @@ public static String asMarkdown(String str) {
         return "```md\n" + str + "```";
     }
 
+    public static String removeMarkdown(String str) {
```
napstr: The method name is incorrect. It does not remove markdown, but escapes it.

FredBoat/src/main/java/fredboat/util/TextUtils.java
```diff
@@ -227,6 +228,25 @@ public static String asMarkdown(String str) {
...
+        ArrayList<Character> markdownList = new ArrayList<Character>() {{
+            add('*');
```
napstr: This list should not be created each time this method is called.

FredBoat/src/main/java/fredboat/util/TextUtils.java
```diff
@@ -227,6 +233,19 @@ public static String asMarkdown(String str) {
         return "```md\n" + str + "```";
     }
 
+    public static String escapeMarkdown(String str) {
+        StringBuilder revisedString = new StringBuilder();
```
Shredder121: Could the StringBuilder have the same initial size as the original String?

BadJukeBox: It could i believe, but I'm unsure of what good that does, can you let me know what you think?

napstr: There is an underlying array for each StringBuilder with a default capacity of 16. The StringBuilder uses this array to store its content. What then happens is if the content length exceeds the underlying array, that the StringBuilder allocates a bigger array and copies the contents over. This process might happen several times.

So if you know the size (or at least the dimension) of the resulting String in advance (which we do here), it is possible to set a custom capacity that is much closer to the expected length than the default capacity, saving those allocations.

BadJukeBox: Ah that makes sense. Do you think it would be better to just use the base string length, or take a pass through the original string and count the number of markdown characters, then allocate the exact amount of space?

Shredder121: No passes, the original size is fine.
It's already requestedSize+16 again, so that should be fine.

napstr:
What's the status here? #375 needs a markdown escaper so I'd like to merge this soon.

FredBoat/src/main/java/fredboat/util/TextUtils.java
```diff
@@ -45,10 +45,19 @@
...
 public class TextUtils {
 
     private static final Pattern TIMESTAMP_PATTERN = Pattern.compile("^(\\d?\\d)(?::([0-5]?\\d))?(?::([0-5]?\\d))?$");
+
+    private static final ArrayList<Character> markdownList = new ArrayList<Character>() {{
+        add('*');
+        add('`');
+        add('~');
+        add('_');
+    }};
```
Shredder121: Please use `Arrays.asList()` for this 😅

FredBoat/src/main/java/fredboat/util/TextUtils.java
```diff
@@ -243,6 +252,19 @@ public static String asCodeBlock(String str, String... style) {
         return "```" + sty + "\n" + str + "\n```";
     }
 
+    public static String escapeMarkdown(String str) {
+        StringBuilder revisedString = new StringBuilder(str.length());
+        for (Character n : str.toCharArray()) {
+            if (markdownList.contains(n)) {
+                revisedString.append("\\" + n);
```
Shredder121: String concatenation in append.
You can `append('\\').append(n)` or leave out the `append(n)`, remove the `else` and let the statement at `261` append the character.

PR #368, opened 2017-10-19
Add Topic Channel Announcement

### LINKED ISSUE #300: Display whats now playing in music channel topic
Something like the picture down below.

schermafbeelding 2017-09-18 om 17 30 12

I think it looks great, you can immediately see what's playing and by who!

Nanabell:
This is the starting to fix #300. I'm not happy with how the topic is enabled but I just put it somewhere to get it working. Currently the only supported topic style is when playing a song I didn't add anything special for pause/stop etc.

### DIFF
--- a/FredBoat/src/main/java/fredboat/audio/player/GuildPlayer.java
+++ b/FredBoat/src/main/java/fredboat/audio/player/GuildPlayer.java
@@ -44,6 +44,8 @@ import fredboat.feature.I18n;
 import fredboat.messaging.CentralMessaging;
 import fredboat.perms.PermissionLevel;
 import fredboat.perms.PermsUtil;
+import fredboat.util.Emojis;
+import fredboat.util.TextUtils;
 import net.dv8tion.jda.core.JDA;
 import net.dv8tion.jda.core.Permission;
 import net.dv8tion.jda.core.entities.Guild;
@@ -51,6 +53,8 @@ import net.dv8tion.jda.core.entities.Member;
 import net.dv8tion.jda.core.entities.TextChannel;
 import net.dv8tion.jda.core.entities.VoiceChannel;
 import net.dv8tion.jda.core.managers.AudioManager;
+import net.dv8tion.jda.core.managers.ChannelManager;
+import net.dv8tion.jda.core.requests.restaction.ChannelAction;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
@@ -97,6 +101,7 @@ public class GuildPlayer extends AbstractPlayer {
         if (getRepeatMode() != RepeatMode.SINGLE && isTrackAnnounceEnabled() && !isPaused()) {
             TextChannel activeTextChannel = getActiveTextChannel();
             if (activeTextChannel != null) {
+                updateTopic();
                 CentralMessaging.sendMessage(activeTextChannel,
                         atc.i18nFormat("trackAnnounce", atc.getEffectiveTitle()));
             }
@@ -459,5 +464,22 @@ public class GuildPlayer extends AbstractPlayer {
 
     private void voteSkipCleanup() {
         VoteSkipCommand.guildSkipVotes.remove(guildId);
+    }
+
+    private void updateTopic() {
+
+        TextChannel channel = getCurrentTC();
+        if (channel != null) {
+            ChannelManager manager = channel.getManager();
+
+            AudioTrackContext playingAtc = getPlayingTrack();
+            String topic = Emojis.PLAY + " **" + playingAtc.getEffectiveTitle() + "** " + playingAtc.getUser().getAsMention() + " [" + TextUtils.formatTime(playingAtc.getEffectiveDuration()) + "]";
+
+            try {
+                manager.setTopic(topic).queue();
+            } catch (Exception ex) {
+                log.warn("Failed to set Topic");
+            }
+        }
     }
 }
--- a/FredBoat/src/main/java/fredboat/audio/player/GuildPlayer.java
+++ b/FredBoat/src/main/java/fredboat/audio/player/GuildPlayer.java
@@ -51,7 +51,6 @@ import net.dv8tion.jda.core.Permission;
 import net.dv8tion.jda.core.entities.*;
 import net.dv8tion.jda.core.managers.AudioManager;
 import net.dv8tion.jda.core.managers.ChannelManager;
-import net.dv8tion.jda.core.requests.restaction.ChannelAction;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
@@ -59,7 +58,6 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import javax.xml.soap.Text;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -471,8 +469,6 @@ public class GuildPlayer extends AbstractPlayer {
         TextChannel topicChannel = guild.getTextChannelById(gc.getTopicChannel());
 
         if (topicChannel != null) {
-
-
             ChannelManager manager = topicChannel.getManager();
 
             AudioTrackContext playingAtc = getPlayingTrack();
@@ -486,7 +482,5 @@ public class GuildPlayer extends AbstractPlayer {
                 log.warn("Failed to set Topic");
             }
         }
-
-
     }
 }
--- a/FredBoat/src/main/java/fredboat/audio/player/GuildPlayer.java
+++ b/FredBoat/src/main/java/fredboat/audio/player/GuildPlayer.java
@@ -94,12 +94,24 @@ public class GuildPlayer extends AbstractPlayer {
     }
 
     private void announceTrack(AudioTrackContext atc) {
-        if (getRepeatMode() != RepeatMode.SINGLE && isTrackAnnounceEnabled() && !isPaused()) {
-            TextChannel activeTextChannel = getActiveTextChannel();
-            if (activeTextChannel != null) {
-                updateTopic();
-                CentralMessaging.sendMessage(activeTextChannel,
-                        atc.i18nFormat("trackAnnounce", atc.getEffectiveTitle()));
+        Guild guild = getGuild();
+        if (guild == null)
+            return;
+
+        GuildConfig gc = EntityReader.getGuildConfig(guild.getId());
+        if (getRepeatMode() != RepeatMode.SINGLE && !isPaused()) {
+
+            if (isTrackAnnounceEnabled(gc)) {
+                TextChannel activeTextChannel = getActiveTextChannel();
+                if (activeTextChannel != null) {
+                    CentralMessaging.sendMessage(activeTextChannel,
+                            atc.i18nFormat("trackAnnounce", atc.getEffectiveTitle()));
+                }
+            }
+
+            TextChannel topicChannel = getTopicChannel(gc);
+            if (topicChannel != null) {
+                updateTopic(guild, topicChannel);
             }
         }
     }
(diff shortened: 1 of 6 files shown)


FredBoat/src/main/java/fredboat/audio/player/GuildPlayer.java
```diff
@@ -44,20 +44,22 @@
...
+import net.dv8tion.jda.core.managers.ChannelManager;
+import net.dv8tion.jda.core.requests.restaction.ChannelAction;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import javax.xml.soap.Text;
```
napstr: This looks out of place.

FredBoat/src/main/java/fredboat/util/ArgumentUtil.java
```diff
@@ -26,13 +26,10 @@
...
 import fredboat.commandmeta.abs.CommandContext;
-import net.dv8tion.jda.core.entities.Guild;
-import net.dv8tion.jda.core.entities.IMentionable;
-import net.dv8tion.jda.core.entities.Member;
-import net.dv8tion.jda.core.entities.Message;
-import net.dv8tion.jda.core.entities.Role;
+import net.dv8tion.jda.core.entities.*;
 
 import javax.annotation.Nonnull;
+import javax.xml.soap.Text;
```
napstr: This looks out of place.

FredBoat/src/main/java/fredboat/audio/player/GuildPlayer.java
```diff
@@ -97,6 +99,7 @@ private void announceTrack(AudioTrackContext atc) {
         if (getRepeatMode() != RepeatMode.SINGLE && isTrackAnnounceEnabled() && !isPaused()) {
             TextChannel activeTextChannel = getActiveTextChannel();
             if (activeTextChannel != null) {
+                updateTopic();
```
napstr: This should be fine to call outside of the textchannel specific null check.

FredBoat/src/main/java/fredboat/audio/player/GuildPlayer.java
```diff
@@ -460,4 +463,30 @@ void destroy() {
...
+
+    private void updateTopic() {
+        Guild guild = getGuild();
+
+        GuildConfig gc = EntityReader.getGuildConfig(guild.getId());
+        TextChannel topicChannel = guild.getTextChannelById(gc.getTopicChannel());
+
+        if (topicChannel != null) {
+
+
```
napstr: Whats up with the excessive whitespace here?

FredBoat/src/main/java/fredboat/audio/player/GuildPlayer.java
```diff
@@ -460,4 +463,30 @@ void destroy() {
...
+            String topic = Emojis.PLAY + " **" + playingAtc.getEffectiveTitle() + "** " + user.getAsMention() + " [" + TextUtils.formatTime(playingAtc.getEffectiveDuration()) + "]";
+
+            try {
+                manager.setTopic(topic).queue();
+            } catch (Exception ex) { //dunno what Exception is thrown when missing perms so i just catch all
+                log.warn("Failed to set Topic");
+            }
+        }
+
+
```
napstr: Whats up with the excessive whitespace here?

FredBoat/src/main/java/fredboat/command/moderation/ConfigCommand.java
```diff
@@ -55,11 +57,13 @@ public void onInvoke(@Nonnull CommandContext context) {
 
     private void printConfig(CommandContext context) {
         GuildConfig gc = EntityReader.getGuildConfig(context.guild.getId());
+        TextChannel channel = context.guild.getTextChannelById(gc.getTopicChannel());
 
         MessageBuilder mb = CentralMessaging.getClearThreadLocalMessageBuilder()
                 .append(context.i18nFormat("configNoArgs", context.guild.getName())).append("\n")
                 .append("track_announce = ").append(gc.isTrackAnnounce()).append("\n")
                 .append("auto_resume = ").append(gc.isAutoResume()).append("\n")
+                .append("topic_channel = ").append(channel != null ? channel.getName() : "Null").append("\n")
```
napstr: 1. "Null" is not an end-user friendly word. "None" would be better imo.
2. topic_channel feels like a suboptimal end-user facing name for the config option because it isn't really self explanatory. I don't have a better suggestion though, sorry.

FredBoat/src/main/java/fredboat/db/entity/GuildConfig.java
```diff
@@ -55,6 +55,9 @@
     @Column(name = "lang", nullable = false)
     private String lang = "en_US";
 
+    @Column(name = "topic_channel", nullable = false)
+    private String topicChannel = "";
```
napstr: why use string when snowflakes can be longs?

FredBoat/src/main/java/fredboat/db/entity/GuildConfig.java
```diff
@@ -55,6 +55,9 @@
     @Column(name = "lang", nullable = false)
     private String lang = "en_US";
 
+    @Column(name = "topic_channel", nullable = false)
```
napstr: `nullable = false` breaks hibernate auto ddl and the whole entity, because it is adding a non-null column without a default value to existing entries.
Several options to handle the issue here:
- write migration queries that have to be run manually before deploying this branch setting a default non-null value (meh solution imo)
- make it nullable and handle null values in the code
- use the `ColumnDefault` annotation (see BlacklistEntry.java for an example usage) and handle the chosen default value in the code

FredBoat/src/main/java/fredboat/audio/player/GuildPlayer.java
Nanabell: LOL tbh no idea how that got there

FredBoat/src/main/java/fredboat/audio/player/GuildPlayer.java
Nanabell: ref op ^
(just didn't bother to move it yet was from when it was bound to active tc)

FredBoat/src/main/java/fredboat/command/moderation/ConfigCommand.java
Nanabell: make make this whole thing a separate command?

FredBoat/src/main/java/fredboat/db/entity/GuildConfig.java
Nanabell: Oriented myself after https://github.com/Frederikam/FredBoat/blob/master/FredBoat/src/main/java/fredboat/db/entity/GuildPermissions.java#L66-L72 but sure will change it

FredBoat/src/main/java/fredboat/command/moderation/ConfigCommand.java
napstr: Idk. Up to you, I'm fine with this sitting a bit hidden in the ;;config for now.

FredBoat/src/main/java/fredboat/db/entity/GuildConfig.java
napstr: That's a list of ids tho. This is just a single one.

FredBoat/src/main/java/fredboat/db/entity/GuildConfig.java
Nanabell: cant I just use longs and just do topicChannel = 0; 0 is not null & cant belong to a guild

napstr: The issue is that primitives can't be null. By adding that column to existing rows without a default value, the value in the database is null. Hibernate then throws an NPE when you load such an entity. This is a rather sneaky exception and easy to miss when working with throwaway local dbs.
One solution is to not use primitives (Long instead of long) and have null checks in place, or use the `@ColumnDefault` annotation to have auto-ddl set those default values.

Nanabell: ah, I see. misunderstood the "default" value

FredBoat/src/main/java/fredboat/db/entity/GuildConfig.java
```diff
@@ -55,6 +56,10 @@
...
+    @ColumnDefault("0")
+    private Long topicChannel = 0L;
```
napstr: Why use the non-primitive long here? If you are defining the column properly (which you are) you can use the primitive as it implicitly guarantees to never be null and therefore saves you a null check.
Using a Long makes sense when you are expecting a column to contain null values.

FredBoat/src/main/java/fredboat/audio/player/GuildPlayer.java
```diff
@@ -435,16 +448,27 @@ public void onTrackStart(AudioPlayer player, AudioTrack track) {
...
 
+    @Nullable
+    private TextChannel getTopicChannel(GuildConfig gc) {
+        TextChannel channel = null;
+        try {
+            Guild guild = getGuild();
+            if (guild != null) {
+                channel = getGuild().getTextChannelById(gc.getTopicChannel());
+            }
+        } catch (DatabaseNotReadyException ignored) {}
```
napstr: This Exception can never occur in this code at this place.

FredBoat/src/main/java/fredboat/audio/player/GuildPlayer.java
```diff
@@ -435,16 +448,27 @@ public void onTrackStart(AudioPlayer player, AudioTrack track) {
         super.onTrackStart(player, track);
     }
 
-    private boolean isTrackAnnounceEnabled() {
+    private boolean isTrackAnnounceEnabled(GuildConfig config) {
         boolean enabled = false;
         try {
-            GuildConfig config = EntityReader.getGuildConfig(Long.toString(guildId));
             enabled = config.isTrackAnnounce();
         } catch (DatabaseNotReadyException ignored) {}
```
napstr: This Exception can never occur in this code at this place.

FredBoat/src/main/java/fredboat/audio/player/GuildPlayer.java
```diff
@@ -94,11 +94,24 @@ public GuildPlayer(Guild guild) {
...
-        if (getRepeatMode() != RepeatMode.SINGLE && isTrackAnnounceEnabled() && !isPaused()) {
-            TextChannel activeTextChannel = getActiveTextChannel();
-            if (activeTextChannel != null) {
-                CentralMessaging.sendMessage(activeTextChannel,
-                        atc.i18nFormat("trackAnnounce", atc.getEffectiveTitle()));
+        Guild guild = getGuild();
+        if (guild == null)
+            return;
+
+        GuildConfig gc = EntityReader.getGuildConfig(guild.getId());
```
napstr: This is the place to catch the DatabaseException.

FredBoat/src/main/java/fredboat/command/moderation/ConfigCommand.java
```diff
@@ -103,6 +107,15 @@ private void setConfig(CommandContext context) {
                     context.reply(context.i18nFormat("configMustBeBoolean", invoker.getEffectiveName()));
                 }
                 break;
+            case "topic_channel":
```
napstr: How to unset this?

FredBoat/src/main/java/fredboat/audio/player/GuildPlayer.java
```diff
@@ -460,4 +484,21 @@ void destroy() {
...
+            ChannelManager manager = topicChannel.getManager();
+
+            AudioTrackContext playingAtc = getPlayingTrack();
+            User user = playingAtc.getUser() != null ? playingAtc.getUser() : guild.getSelfMember().getUser();
+
+            String topic = Emojis.PLAY + " **" + playingAtc.getEffectiveTitle() + "** " + user.getAsMention() + " [" + TextUtils.formatTime(playingAtc.getEffectiveDuration()) + "]";
+
+            try {
+                manager.setTopic(topic).queue();
+            } catch (Exception ex) { //dunno what Exception is thrown when missing perms so i just catch all
```
napstr: Read the docs of that method. Also a log inside a catch, that doesn't log the thrown exception, is 100% useless.
This whole PR is missing feedback to the user in case FredBoat is lacking the needed permissions.
Also general documentation of the feature is missing. This will lead barely anyone using the feature, those that do use it will be unhappy with the lacking feedback of why it is not working, and unnecessary load in #support.

Macley-Kun:
Is there an update on this pull?

Raven-Rose:
I'm assuming something like this will more than likely wait until after the rewrite

Macley-Kun:
Sure no problem! Was just wondering/worried if there would come an update. Thank you for your answer!

PR #369, opened 2017-10-19
Add mention prefix, rework command argument parsing

napstr:
First step to custom prefixes is to make FredBoat reliably answer when mentioned - that's what this PR does.
Cleaned up the parsing of arguments - they are really just the arguments now, split by whitespace, no prefix or commands to be handled in args[0] etc. That's why a lot of commands have been touched in this PR, but that's mostly just refactoring.
Important changes are found in EventListenerBoat and CommandContext.

### DIFF
--- a/FredBoat/src/main/java/fredboat/FredBoat.java
+++ b/FredBoat/src/main/java/fredboat/FredBoat.java
@@ -51,6 +51,7 @@ import net.dv8tion.jda.core.JDA;
 import net.dv8tion.jda.core.JDAInfo;
 import net.dv8tion.jda.core.entities.Guild;
 import net.dv8tion.jda.core.entities.TextChannel;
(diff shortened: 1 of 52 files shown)


FredBoat/src/main/java/fredboat/FredBoat.java
```diff
@@ -374,6 +375,15 @@ public static Guild getGuildById(long id) {
         return null;
     }
 
+    @Nullable
+    public static User getUserById(long id) {
```
Shredder121: Method is never used?

napstr: Needed it for some iteration, decided to leave it in together with the other global lookups /shrug

napstr:
Thanks.

PR #372, opened 2017-10-21
Feature/player limitations

Nanabell:
This pr will allow guilds to limit their players in multiple ways.
Currently supported: 
- Disallow playlists
- limit per user songs (i.e. how many songs a user can queue)
- limit song length

at the moment `allow_playlist` is only limited to users with `USER` permission as `DJ` bypasses this. 
Should this be the case for all of these checks or should they be enforced for everyone? I was indecisive on this one so I want your feedback b4 this can be merged


FredBoat/src/main/java/fredboat/db/entity/GuildConfig.java
```diff
@@ -55,6 +55,15 @@
     @Column(name = "lang", nullable = false)
     private String lang = "en_US";
 
+    @Column(name = "allow_playlist", nullable = false)
```
napstr: This breaks hibernate ddl / the entity. The entity will throw an NPE when an existing entity is loaded from a row that doesn't have any value set for this column.
See https://github.com/Frederikam/FredBoat/pull/368#discussion_r146131750 for more details on a solution.

Please apply a solution to the other two values below too.

FredBoat/src/main/java/fredboat/audio/queue/AudioLoader.java
```diff
@@ -235,8 +319,8 @@ public void loadFailed(FriendlyException fe) {
         loadNextAsync();
     }
 
-    private void loadSplit(AudioTrack at, IdentifierContext ic){
-        if(!(at instanceof YoutubeAudioTrack)){
+    private void loadSplit(AudioTrack at, IdentifierContext ic) {
+        if (!(at instanceof YoutubeAudioTrack)) {
```
napstr: pl0x no reformatting of lines you aren't touching (check this whole file)

Nanabell: mb~ #AutoFormatting

napstr: There are probably other ways to make this not happen, here's what I use:
image
This setting in the SaveActions plugin is great. Or manually staging the hunks with gitkraken and discarding those where the formatter accidently ran on.

FredBoat/src/main/java/fredboat/audio/queue/AudioLoader.java
```diff
@@ -258,7 +342,7 @@ private void loadSplit(AudioTrack at, IdentifierContext ic){
 
             String title1 = m.group(1);
             String title2 = m.group(3);
-            
+
```
Nanabell: fk this one, this is staying. I cant save w/o IntelliJ removing it

FredBoat/src/main/java/fredboat/audio/player/GuildPlayer.java
```diff
@@ -219,6 +219,22 @@ public int getTrackCount() {
         return result;
     }
 
+    public List<AudioTrackContext> getAllTracks() {
```
napstr: It is a deliberate decision to not have such a method in the TrackProvider, because it will be very inefficient once tracks are persistent. Please to not introduce it here again, especially not in such hacky way.

FredBoat/src/main/java/fredboat/audio/queue/AudioLoader.java
```diff
@@ -166,14 +170,33 @@ private PlaylistInfo getSlowLoadingPlaylistData(String identifier) {
 
     @Override
     public void trackLoaded(AudioTrack at) {
+        GuildConfig gc = EntityReader.getGuildConfig(context.getGuild().getId());
```
napstr: What happens if the DB fails?

FredBoat/src/main/java/fredboat/audio/queue/AudioLoader.java
```diff
@@ -196,6 +219,21 @@ public void trackLoaded(AudioTrack at) {
 
     @Override
     public void playlistLoaded(AudioPlaylist ap) {
+        GuildConfig gc = EntityReader.getGuildConfig(context.getGuild().getId());
```
napstr: What happens if the DB fails?

FredBoat/src/main/java/fredboat/command/moderation/ConfigCommand.java
```diff
@@ -99,6 +103,30 @@ private void setConfig(CommandContext context) {
...
+                if (val.matches("^\\d+$")) {
+                    gc.setMemberTrackLimit(Integer.valueOf(val));
+                    EntityWriter.mergeGuildConfig(gc);
+                    context.replyWithName("`member_track_limit` " + context.i18nFormat("configSetTo", val));
+                } else {
+                    context.replyWithName(context.i18n("configMustBeNumber"));
+                }
+                break;
+            case "max_track_duration":
+                Long time = TextUtils.parseTimeString(val); // We should handle failed parses better than throwing tbh.
```
napstr: > // We should handle failed parses better than throwing tbh

No, catching NumberFormatExceptions is fine. It is not the cleanest solution there is but certainly the less verbose one. And it is missing here.

FredBoat/src/main/java/fredboat/command/moderation/ConfigCommand.java
```diff
@@ -99,6 +103,30 @@ private void setConfig(CommandContext context) {
...
+                    gc.setAllowPlaylist(Boolean.valueOf(val));
+                    EntityWriter.mergeGuildConfig(gc);
+                    context.replyWithName("`allow_playlist` " + context.i18nFormat("configSetTo", val));
+                } else {
+                    context.replyWithName(context.i18n("configMustBeBoolean"));
+                }
+                break;
+            case "member_track_limit":
+                if (val.matches("^\\d+$")) {
+                    gc.setMemberTrackLimit(Integer.valueOf(val));
```
napstr: This should be `Integer.parseUnsignedInt()` unless you are expecting negative values.

FredBoat/src/main/java/fredboat/command/moderation/ConfigCommand.java
```diff
@@ -99,6 +103,30 @@ private void setConfig(CommandContext context) {
                     context.reply(context.i18nFormat("configMustBeBoolean", invoker.getEffectiveName()));
                 }
                 break;
+            case "allow_playlist":
+                if (val.equalsIgnoreCase("true") | val.equalsIgnoreCase("false")) {
+                    gc.setAllowPlaylist(Boolean.valueOf(val));
```
napstr: Please use `Boolean.parseBoolean` instead. Also be aware that this returns false for anything that is not "true".

FredBoat/src/main/java/fredboat/audio/queue/AudioLoader.java
```diff
@@ -167,18 +175,58 @@ private PlaylistInfo getSlowLoadingPlaylistData(String identifier) {
...
+                    if (maxDuration > 0) {
+                        if (at.getDuration() > maxDuration) {
+                            context.reply(context.i18nFormat("exceedsTrackDurationLimitSingle",
+                                    "**" + at.getInfo().title + "**",
+                                    "`" + TextUtils.formatTime(maxDuration) + "`"));
+                            return;
+                        }
+                    }
+
+                    // Move Napstes Metrics Magic so it only increments on successful loaded tracks
```
napstr: This comment is unnecessary. Metrics should be pretty much self explanatory. The reason you probably decided to add this is because you are changing kind of the metric collected (tracks added instead of tracks loaded) without changing the name of the metric and its description and help (see the Metrics class for details).
Please don't change the behaviour of existing metrics, as that destroys the meaning of the collected data. Instead add a new one if you think one is needed.

FredBoat/src/main/resources/lang/en_US.properties
```diff
@@ -301,3 +303,8 @@ voteSkipEmbedVoters={0} out of {1} have voted to skip the current track
 mathOperationResult=The result is
 mathOperationDivisionByZeroError=I cannot divide by zero.
 mathOperationInfinity=The number is too big to be displayed!
+cantQueuePlaylist=To queue playlists you need to have {0} permission!
+exceedsMemberTrackLimitMultiple=Failed to queue {0} tracks, you can only have a maximum of {1} tracks in the queue.
+exceedsMemberTrackLimitSingle=Failed to queue {0}, you can only have a maximum of {1} tracks in the queue.
+exceedsTrackDurationLimitSingle=Failed to queue {0}. Track exceeds Duration limit of {1}.
```
napstr: Duration -> duration

Issue #129, opened 2017-04-05
Additional prefix for non-patron bots


Frederikam:
We need an additional prefix that the non-patron bots can respond to, so that patrons can use both bots in the same text channel.

Maybe use `::` as prefix?

Nanabell:
why cant we change the prefix of the patron bot ?

Frederikam:
Because patrons are already accustomed to `;;`. Maybe I can do a poll on Patreon

Nanabell:
hmm even if i would say change the prefix on the smaller scale. Would only cause confusion if suddenly a second prefix would work for the main bot. If we decide to do a second prefix i would suggest adding it to Patron

napstr:
#69 (customizable prefix) should solve this issue too.

napstr:
Funny how this issue is actually fully solved (emphasis on the _non-patron_ part).

PR #375, opened 2017-10-22
Add custom prefix support Resolves #69 & #129

### LINKED ISSUE #69: Suggestion: Prefix editing
would it be possible to allow users to edit the prefix without having to dig into the java files and editing the return messages and the prefix constant?

so for example: if you dont want ;; as your prefix you can set in the config.json to be !

thanks

Also Manage messages, permission is enabled for the bots role but it wont do ;;clear. if you want i can create a second

napstr:
This feature PR allows custom prefixes on a guild scope. Required permission level is admin.

ToDos after review, before merge:
- [x] Resolve escape markdown todos (relies on #367 getting merged)
- [x] Update language translation files


napstr:
Anyone please feel free to play around with this / review the code / check the new translation strings.

napstr:
This PR suffers from one major flaw caused by our current infrastructure setup: Since the public and patron FredBoat are using the same database, it would be impossible to set different prefixes for them.

Some thinking needs to go into what data we want to be available globally, and what data should be bot scoped, and how that should be achieved (different databases? different tables?).

PR #376, opened 2017-10-23
Feature/ff additions

Nanabell:
This Change will allow Selfhosters to turn off the VoiceChannelCleanupAgent regardless of distribution as well as allow them to change if the player should pause when its left alone in the voice channel.

The use cases
If self-hosters want to run the full bot i.e dev distro but don't want to have the vcCleanup running. And we've gotten multiple requests from users to have a "radio" which plays 24/7, while we cant do that due to resources, self-hosters should be able to turn it on for them if they think they want.

This pr should require no change from us on the public bot as it defaults to the current Fredboat settings.
**But** the patron bot will need to have the vcCleanup disabled

### DIFF
--- a/FredBoat/config.yaml
+++ b/FredBoat/config.yaml
@@ -6,6 +6,7 @@ restServerEnabled: true        # Set this to false if you are running multiple F
 admins:            []          # Add comma separated userIds and roleIds that should have access to bot admin commands
 useAutoBlacklist:  true        # Set to true to automatically blacklist users who frequently hit the rate limits
 game:              ""          # Set the displayed game/status. Leave empty quote marks for the default status
+continuePlayback:  false        # Set to true to force the player to continue playback even if left alone
 
 enableYouTube:     true        # Set to true to enable playing YouTube links
 enableSoundCloud:  true        # Set to true to enable playing SoundCloud links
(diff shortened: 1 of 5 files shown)


napstr:
I don't want to have to think about selfhosters when changing around feature flags, nor document their functionality for them.

babdab9:
When will this be added in the stable build?

napstr:
In the future. Maybe.

napstr:
I want to expland on the reasons as my comments from yesterday are barely constructive.

##### Feature Flags
Somewhat contrary to their names they are not actually meant to toggle features, at least not in an enduser / production way. Feature flags are a tool for developers to test things out. Once introduced, they need to be removed as soon as they served their purpose of making sure everything is working as intended.
Related read: http://swreflections.blogspot.de/2014/08/feature-toggles-are-one-of-worst-kinds.html

##### What to use instead
The config already contains all relevant options for selfhosters and is documented adequately. New config options should go in there.

Btw, turning off the VoiceChannelCleanup agent won't be necessary after modularizing the commands, because as soon as that happens, the development distribution can go back to being the internal development distribution and all selfhosters can be advised to use the the Patron distribution which has no VoiceChannelCleanup.

Nanabell:
Should I even bother making the VoiceChannelCleanup a config option then? We have no eta on that

PR #379, opened 2017-10-26
Adding leave server command

bobatrance:
Ref issue #378

Issue #378, opened 2017-10-25
Leave Server command


KKHaneul:
This thread is for suggest.

Only Bot owner can use this command.
At send ;;leaveserver command in server what bot is entered, bot can leave server self.

babdab9:
Excuse me, but because English is not my native toungue, can I have a better explanation :D

RealKC:
@babdab9 OP is asking for a command that makes FredBoat leave the server the command is excuted in.

PR #380, opened 2017-10-26
Adding leave server command

bobatrance:
Ref issue #378, fixed PR to dev branch #379.

### DIFF
--- a/FredBoat/src/main/java/fredboat/command/admin/LeaveServerCommand.java
+++ b/FredBoat/src/main/java/fredboat/command/admin/LeaveServerCommand.java
@@ -31,14 +31,22 @@ import fredboat.commandmeta.abs.CommandContext;
 import fredboat.commandmeta.abs.ICommandRestricted;
 import fredboat.messaging.internal.Context;
 import fredboat.perms.PermissionLevel;
+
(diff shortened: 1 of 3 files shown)


FredBoat/src/main/java/fredboat/command/admin/LeaveServerCommand.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public LeaveServerCommand(String name, String... aliases) {
+        super(name, aliases);
+    }
+
+    @Override
+    public void onInvoke(@Nonnull CommandContext context) {
+
+        try {
+            context.replyWithName("Thanks for having me!").getWithDefaultTimeout();
```
napstr: Why block the main thread here instead of using the success consumer?

bobatrance: I wanted it to match the design pattern of the other admin commands I looked at. I can change this to use a success callback.

napstr: I am not aware of any "block the main thread without a reason" design pattern in FredBoat.

PR #395, opened 2017-11-09
Command modules resolves  #291

### LINKED ISSUE #291: Rework the command system
Goals:
- Commands will be split into 4 groups:
    - Default (includes all music commands, help, etc)
    - Moderation
    - Util
    - Fun
- Groups can be enabled and disabled with the exception of default. The 3 groups are disabled by default
- Groups and permissions will be defined by an optional annotation
- MAIN and MUSIC will merge to one.

napstr:
This PR merges all existing commands (formerly split into music and main bot).

The commands are organized in the following modules:
- Admin: BOT_OWNER and BOT_ADMIN commands
- Info: Debug and Informational commands that show FredBoat related information
- Config: Any commands that allow users to change FredBoat settings
- Music: duh
- Moderation: duh (kicking banning etc)
- Util: anything else, really (other APIs for example)
- Fun: memes games images etc

Admin, Info and Config being more of an internal organization unit, they are not meant to be disabled.

By default the music module is enabled (and locked in as that), the other three modules are disabled.

The information about which module is enabled/disabled is saved with a bitfield as part of the GuildConfig.

#291 was detailing to use annotations to define the modules, but I had a hard time wrapping my head around a decent way to implement that, because we reuse several commands at several places for different purposes. The approach taken in this PR allows for high flexibility in defining what a command module is (just register commands into it).

#### ToDos before merge:
- [x] rework CommandsCommand, MusicHelpCommand and HelpCommand to reflect the modularity
- [x] Update translations files
- [x] revisit how the enabled module are saved in the guildconfig + default value after decision is reached in handling opt-in vs opt-out
- [x] PR changes to docs https://github.com/FredBoat/fredboat.com/pull/17

#### Questions I'd like to discuss, besides general feedback to the code:
- Do we really want to have all modules except music be opt-in? From the data we have about the usage of existing configuration commands, very few users actually bother configuring their FredBoat. Given that enabling/disabling modules is an admin-level command, there is this additional barrier for regular users to get access to the disabled modules. Therefore I'd much prefer an opt-out approach for command modules.
- something else that I forgot lol, will add later if I remember


Nanabell:
tbf after hearing arguments for enabling all modules by default i would vote to have em enabled

babdab9:
When will this be (finally) merged?

Nanabell:
When it's done. If you would read the op it has a section called `ToDos before merge` and it has merge conflicts

babdab9:
hmm :thonk: ok

napstr:
This PR needs to switch to opt-out modules, and for that the persistence needs to be solved differently.
The current, Discord permissions style, might look nifty and concise, but is probably not that great to work with (try answering how many guilds have a specific modules disabled), and hard locks us into a model of opt-in / opt-out due it's nature.

babdab9:
what do you mean by `opt-out`? Also, last thing on the todo list can be done by me
-Player Hwan

napstr:
@Frederikam pl0x review this. Mostly interested in your opinion on the ;;commands and ;;modules output and handling.

napstr:
When testing use a non-admin/non-owner account, since the output varies for us.

Frederikam:
Can you provide screenshots?

napstr:
Here's a bunch.

image
image
image
image
image

napstr:
I'd like to preemptively point out that using Embeds reduces the font size, meaning we can put a lot more things in a smaller space.

FredBoat/src/main/java/fredboat/command/info/MusicHelpCommand.java
```diff
@@ -163,6 +163,7 @@ private static void updateMessage(CommandContext context) {
                 .filter(command -> !(command instanceof PlayCommand
                         && (command.name.equals(CommandInitializer.YOUTUBE_COMM_NAME)
                         || command.name.equals(CommandInitializer.SOUNDCLOUD_COMM_NAME))))
+                .filter(command -> !(command instanceof DestroyCommand))
```
Frederikam: This looks cheap to me. I was suggesting that you would create a generic method in the Command class that returns true if the command should be hidden. That allows us to dynamically hide other commands.

napstr: I think it's too early to build an abstraction for that: We have one general case (DestroyCommand) and one special case (Play command clones yt/sc only, but only hidden in the long music comm list). I wouldn't even know what and how to abstract this in a nice way so it's actually useful and doesn't get in the way. The work can be done when it's actually needed, not for the sake of having it but not using it, and shouldn't take any longer.

Frederikam: Fair enough

napstr:
Documentation PR is up at https://github.com/FredBoat/fredboat.com/pull/17

napstr:
Wew. A new kind of fun. Merge conflicts in translation files.

Frederikam:
I'm personally looking forward to the millions of conflicts that this will inflict on https://github.com/Frederikam/FredBoat/pull/428

napstr:
herewego.gif

babdab9:
Is it up on the public fred? Wasnt able to check disord theese days

napstr:
Not yet.

PR #398, opened 2017-11-13
Small fix at en_US.properties

knight-ryu12:
I found some problem in language, so I've fixed it.


napstr:
Why is this change to the language files needed? The strings display fine for me:

image

knight-ryu12:
Mobile issue.
Seems it won't actually markdown it.

2017/11/14 午後7:01 "Napster" <notifications@github.com>:

Kurozume:
For reference, on mobile it looks like this

Markdown on mobile


_Sent from Xperia XZ F8331 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

Nanabell:
I'm fine with changing this as you cant actually use markdown in titles and the client just eats it

knight-ryu12:
Pretty much, I know client eats markdown and do absolutely nothing, But in
mobile, they don't.
So I just remove em to actually show better on mobile

2017/11/14 午後9:28 "Weiss Schnee" <notifications@github.com>:

PR #410, opened 2017-12-04
Fix for being stuck in catastrophic backtracking Resolves #399

### LINKED ISSUE #399: Thread stuck in split select detection regex
We recently encountered an unresponsive shard with its main thread stuck in a regex. Here's the relevant part of the threaddump:
https://hastebin.com/voqugexodu.txt

Shredder121:
Fixes #399 
So instead of using a regex, I propose a constant-time Splitter based approach.
(The `COMMA` splitter should also be reusable for other things)

### DIFF
--- a/FredBoat/src/main/java/fredboat/command/music/control/SelectCommand.java
+++ b/FredBoat/src/main/java/fredboat/command/music/control/SelectCommand.java
@@ -80,14 +80,7 @@ public class SelectCommand extends Command implements IMusicCommand, ICommandRes
             if (StringUtils.isNumeric(commandOptions)) {
                 requestChoices.add(Integer.valueOf(commandOptions));
             } else if (TextUtils.isSplitSelect(sanitizedQuery)) {
-                // Remove all non comma or number characters
-                String[] querySplit = sanitizedQuery.split(",|\\s");
-
-                for (String value : querySplit) {
-                    if (StringUtils.isNumeric(value)) {
-                        requestChoices.add(Integer.valueOf(value));
-                    }
-                }
+                requestChoices.addAll(TextUtils.getSplitSelect(sanitizedQuery));
             }
 
             //Step 2: Use only valid numbers (usually 1-5)
--- a/FredBoat/src/main/java/fredboat/command/music/control/SelectCommand.java
+++ b/FredBoat/src/main/java/fredboat/command/music/control/SelectCommand.java
@@ -75,12 +75,11 @@ public class SelectCommand extends Command implements IMusicCommand, ICommandRes
 
             // Combine all args and the command trigger. if the trigger is not a number it will be sanitized away
             String commandOptions = (context.trigger + " " + context.rawArgs).trim();
-            String sanitizedQuery = sanitizeQueryForMultiSelect(commandOptions).trim();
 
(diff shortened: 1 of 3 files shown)


napstr:
Does this work on whitespace? Current behavior does + I find it much more convenient myself :upside_down_face:

FredBoat/src/main/java/fredboat/util/TextUtils.java
```diff
@@ -44,16 +47,23 @@
...
 
 public class TextUtils {
 
     private static final Pattern TIMESTAMP_PATTERN = Pattern.compile("^(\\d?\\d)(?::([0-5]?\\d))?(?::([0-5]?\\d))?$");
 
     private static final List<Character> markdownChars = Arrays.asList('*', '`', '~', '_');
 
+    public static final Splitter COMMA = Splitter.on(',')
+            .omitEmptyStrings() // 1,,2 doesn't sound right
+            .trimResults();// have it nice and trim
```
Shredder121: @napstr This does the whitespace handling.
it tokenizes on comma, so`1, 2, 3` => `1`,`<sp>2`,`<sp>3` get trimmed individually to `1`,`2`,`3`.

Shredder121: O wait I c

FredBoat/src/main/java/fredboat/command/music/control/SelectCommand.java
```diff
@@ -80,14 +80,7 @@ static void select(CommandContext context) {
...
             } else if (TextUtils.isSplitSelect(sanitizedQuery)) {
-                // Remove all non comma or number characters
-                String[] querySplit = sanitizedQuery.split(",|\\s");
-
-                for (String value : querySplit) {
-                    if (StringUtils.isNumeric(value)) {
-                        requestChoices.add(Integer.valueOf(value));
-                    }
-                }
+                requestChoices.addAll(TextUtils.getSplitSelect(sanitizedQuery));
```
napstr: The sanitized query (see method at the bottom of the SelectCommand class) needs to be adjusted for the rules of the matcher (specifically whitespaces), possibly even replaced/merged with it so there is only one place responsible.

Shredder121:
```java
@Test
void nonDigitsInSplitSelect() {
    Assertions.assertIterableEquals(
            Collections.singleton(1),
            TextUtils.getSplitSelect("We are number 1 but even this string matches")
    );
}
```
So are you saying this should match as well?
So when cleaning it up reduces it to just numbers, anything can be made a split select string?

Shredder121: I changed this to be heuristic based, if there is an awful lot of cleanup before it can be used as a split selection, it's not a split selection.
How does that sound?

Shredder121:
In the selectcommand it filters out the numbers, didn't know if it needed to be there still.
Can also move that into the isSplitSelect check for more cleanup :clap:
So shall I do that?

napstr:
If you think that's in scope for this PR sure. It's why I didn't merge right away, just in case it needs more than a line change.

PR #411, opened 2017-12-06
Fix condition check

knight-ryu12:
this if wasn't working correctly...
programmer's nightmare 1: "if" was wrong


knight-ryu12:
what, I've seen something wrong

Frederikam:
wew lad :stuck_out_tongue:

knight-ryu12:
yes it's rad PR

PR #412, opened 2017-12-06
Fix for #407

babdab9:
This PR fixes issue #407

Issue #407, opened 2017-11-29
Remove references to decommissioned FredBoat


napstr:
The decommissioned FredBoat (the one without the musical notes in the name) shows up in several places in the code / documentation still.
Places to check: 
- The help message users receive when saying `;;help`
- The website  (see repo here https://github.com/FredBoat/fredboat.com)
- Maybe others?

babdab9:
Isn't this a string file problem

Frederikam:
Yes

babdab9:
I am getting this, if you don't mind

babdab9:
Website PR should be up soon:tm: Sorry for the waiting but real life sometimes...

babdab9:
aaaaaaaaaaaaand done

babdab9:
🤔 this is still not colsed. Have I done smth wrong?

napstr:
There are some code references still there but those will probably disappear as part of #395 or the aftermath of it. I'm not aware of any place facing users that reference the old boat so I'll consider this issue done.

babdab9:
👍

PR #413, opened 2017-12-06
Resolvation for #407

babdab9:
This is a repost of #412, after Fred's request


napstr:
Going to piggyback on this change that will require the helpDM string to be retranslated and move the invite links to code so that they can be changed in the future if need be without invalidating translations.

babdab9:
:man_shrugging: dunno

Issue #197, opened 2017-06-23
Wrong/Weird lang help text


Jannify:
Discord doesn't show the underline from the lang codes...

![grafik](https://user-images.githubusercontent.com/23176718/27480728-7b3b1fca-5819-11e7-852b-1574d736904f.png)

napstr:
*Better Discord doesn't. Haven't seen this reported for the vanilla client yet.

Frederikam:
It works for me and I am using the vanilla client 
![image](https://user-images.githubusercontent.com/2582617/27481236-e97f6e8a-581b-11e7-8a01-3e1ec43812fa.png)

napstr:
I wouldn't be opposed to changing it a bit around. This issue came up more than once, many ppl appear to be using Better Discord/custom themes. Could show a number next to the languages that can be used? Or format them with *s, not with `s.

Frederikam:
Unfortunately, that would lose the monospace grid and make the list less readable as a result

RealKC:
I just encountered the bug with the vanilla client(Canary, but I doubt that makes a difference)

![img](https://fred.moe/JE-.png)

napstr:
Is putting it into a code block an option?

![image](https://user-images.githubusercontent.com/6048348/29681497-220b66ec-8908-11e7-8095-57bf14649dc6.png)

Kurozume:
I don't think this is betterdiscord related at all
[BetterDiscord](https://Blake.s-ul.eu/I8JpqYi8)

However I do like the suggestion @napstr put through, maybe use asciidoc to differentiate the lang codes from the names
[AsciiDoc formatting](https://Blake.s-ul.eu/jLQxEoR5)

napstr:
It's probably due to a weird combination of font, font-size and zoom level.

Here's how asciidoc looks in vanilla:
![image](https://user-images.githubusercontent.com/6048348/29681989-c110e59a-8909-11e7-8f72-78f859ea4545.png)

Kurozume:
I like how asciidoc looks ~~if only it could be blue not orange~~ but idk what everyone else thinks of it

Kurozume:
We should just embed everything cause [embed master race](https://Blake.s-ul.eu/GJag7VHc) 😛

napstr:
![image](https://user-images.githubusercontent.com/6048348/29682232-95f10416-890a-11e7-8855-6d58d0804df5.png)

What happened there lol?

[...]

PR #414, opened 2017-12-06
Reformat the showing of the guild prefix

Kurozume:
This is just a small PR to adjust the way the prefix is displayed when mentioning the bot or using the `;;prefix` command with no arguments.
Previously, the bot would return the following output:
mscmsgl

However this PR would change the output to the following:
mscmsgl

I believe that this makes the prefix much more readable


napstr:
Ever since the issues with getting the language command to be properly formatted (see #197 for still ongoing stuff) I've tried to avoid using the \` \` markdown.
That is the reason why the current formatting with \*\* \*\* was chosen. Unlike the code markdown, I'm not aware of rendering issues with it.
Not saying it's the best and not saying using the code markdown is out of the question but are there concrete issues with the current approach? Remember that its most important task is to display custom input, not the ;; one, which is available from the bot's status too.

napstr:
Fred ok'd this. Adapted changes via 4031b540add356a9269ad783892d03a2f16e3fdf since source repo has been removed.

PR #421, opened 2017-12-19
Split main and cache db

napstr:
This PR's main goal is to split off our data database from the cache (of search results), by using two database connections.
These two uses of a database are very different in nature, and given the enormous volume that the cache database aggregates over time, I prefer them to be separated.
To my best knowledge this is fully backwards compatible with existing docker and non-docker selfhosters and our own files, and actually fixes some docker issues.
I'm a lazy bastard so I used my [SqlSauce lib](https://github.com/napstr/SqlSauce) for the connections., it's very similar to the existing connection solution.

- [x] manually publish updated postgres image


napstr:
When merging this, an updated postgres docker image needs to be published (I'll take care of that).

Frederikam:
What happens when we attempt to do something that typically uses caching, but we don't have a caching database?

napstr:
The caching db is currently as required as the main db - this PR does the bare minimum of splitting them up. Future commits might change that, possible future changes include gracefully ignoring a not provided / not available caching db.

Frederikam:
I'd prefer that the bot wouldn't explode if the caching DB is missing. I'm thinking more in terms of robustness here.

napstr:
It already explodes if the current db is missing / becomes unavailable.
This PR defaults the caching db to the main db, resulting in similar behaviour.

Sure, we could pack this PR with more features and changes, but then it just gets bloated and not reviewed.

Frederikam:
Fair enough. I skimmed the code and it LGTM. I'll leave you to merge this PR

knight-ryu12:
So are we going to split cache and main DB? LGTM

napstr:
Planning to merge this tomorrow morning/noon.

PR #429, opened 2018-01-02
Add PREFIX_COMM_NAME to fix #423

### LINKED ISSUE #423: ;;prefix should have special status to always work, even if the guild-specific prefix was changed
;;prefix should always be able to be called with the default prefix like with ;;help
image
Consider it done

knight-ryu12:
This'll work, because help command worked like this.

### DIFF
--- a/FredBoat/src/main/java/fredboat/commandmeta/abs/CommandContext.java
+++ b/FredBoat/src/main/java/fredboat/commandmeta/abs/CommandContext.java
@@ -92,7 +92,6 @@ public class CommandContext extends Context {
             input = mentionMatcher.group(3).trim();
             isMention = true;
         }
-
         // or starts with a custom/default prefix
         else {
             String prefix = PrefixCommand.giefPrefix(event.getGuild());
@@ -105,7 +104,8 @@ public class CommandContext extends Context {
                 }
             } else {
                 //hardcoded check for the help or prefix command that is always displayed as FredBoat status
-                if (raw.startsWith(Config.CONFIG.getPrefix() + CommandInitializer.HELP_COMM_NAME) || raw.startsWith(Config.CONFIG.getPrefix() + CommandInitializer.PREFIX_COMM_NAME)) {
+                if (raw.startsWith(Config.CONFIG.getPrefix() + CommandInitializer.HELP_COMM_NAME)
+                        || raw.startsWith(Config.CONFIG.getPrefix() + CommandInitializer.PREFIX_COMM_NAME)) {
                     Metrics.prefixParsed.labels("default").inc();
                     input = raw.substring(Config.CONFIG.getPrefix().length());
                 } else {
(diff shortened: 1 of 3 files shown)


FredBoat/src/main/java/fredboat/commandmeta/abs/CommandContext.java
```diff
@@ -107,6 +108,9 @@ public static CommandContext parse(MessageReceivedEvent event) {
                 if (raw.startsWith(Config.CONFIG.getPrefix() + CommandInitializer.HELP_COMM_NAME)) {
                     Metrics.prefixParsed.labels("default").inc();
                     input = raw.substring(Config.CONFIG.getPrefix().length());
+                } else if (raw.startsWith(Config.CONFIG.getPrefix() + CommandInitializer.PREFIX_COMM_NAME)){
+                    Metrics.prefixParsed.labels("default").inc();
+                    input = raw.substring(Config.CONFIG.getPrefix().length());
```
Shredder121: These two `if` bodies have the same content.
We're not going to be copy pasting fixes around, you can add a `||` after the original if, to check if it's either `help` or `prefix`.
Since that's only what it needs to do, the body already does the heavy lifting.

knight-ryu12: OK. I'll change it.

knight-ryu12:
If you have further review or changes that need to make, please comment 
~~Idk what to do here~~

build.gradle
```diff
@@ -18,6 +18,7 @@ subprojects {
 
     compileJava.dependsOn 'clean'
     compileJava.options.encoding = 'UTF-8'
+
```
napstr: Why this?

FredBoat/src/main/java/fredboat/commandmeta/abs/CommandContext.java
```diff
@@ -92,6 +92,7 @@ public static CommandContext parse(MessageReceivedEvent event) {
             input = mentionMatcher.group(3).trim();
             isMention = true;
         }
+
```
napstr: Why this?

FredBoat/src/main/java/fredboat/commandmeta/abs/CommandContext.java
```diff
@@ -103,8 +104,8 @@ public static CommandContext parse(MessageReceivedEvent event) {
                     Metrics.prefixParsed.labels("custom").inc();
                 }
             } else {
-                //hardcoded check for the help command that is always displayed as FredBoat status
-                if (raw.startsWith(Config.CONFIG.getPrefix() + CommandInitializer.HELP_COMM_NAME)) {
+                //hardcoded check for the help or prefix command that is always displayed as FredBoat status
+                if (raw.startsWith(Config.CONFIG.getPrefix() + CommandInitializer.HELP_COMM_NAME) || raw.startsWith(Config.CONFIG.getPrefix() + CommandInitializer.PREFIX_COMM_NAME)) {
```
napstr: Please add a new line before the ||

PR #436, opened 2018-01-15
Rework EntityReader/Writer for convenience and exception safety

napstr:
Makes use of the database wrapper; wraps all calls to it to catch DatabaseExceptions
Adjusts existing entity classes for that, resolves a todo in the SearchResult.
Added a bunch of nullability tags to stuff I had to touch
Also wrote some docs in the EntityIO class for how to handle things.
 
Tested the database calls, they work fine.
Opening the PR to let you chip in on this.


Frederikam:
Would have been nice if you based this on #428 to avoid merge conflicts.

napstr:
My rebases are battle proofed :kappa:

PR #439, opened 2018-01-23
Send hello message on joining a guild

napstr:
to be tested and pulled in after #436


FredBoat/src/main/java/fredboat/commandmeta/CommandInitializer.java
```diff
@@ -91,6 +94,7 @@ public static void initCommands() {
         infoModule.registerCommand(new FuzzyUserSearchCommand("fuzzy"));
         infoModule.registerCommand(new GetIdCommand("getid"));
         infoModule.registerCommand(new GitInfoCommand("gitinfo", "git"));
+        infoModule.registerCommand(new HelloCommand("hello"));
```
Nanabell: Maybe add ;;welcome as an alias?

napstr: We already have a "welcome" alias in use, as @RealKC noticed, will use the `;;about` one then.

Issue #168, opened 2017-04-29
[suggestion] Let users create their own aliases with ;;alias


LEADER01:
Not sure how you'd implement this for non self hosted bot but this would be really usefully to have

Frederikam:
Can you elaborate? Do you mean allowing server admins to create aliases, or allowing selfhosters to configure a list of aliases?

LEADER01:
Both ideally, but I was thinking about first option - allowing admins to do ;;alias <command> <alias>

Frederikam:
Why both? Wouldn't the first just make the second mostly obsolete?

LEADER01:
*looking away ashamed of own stupidity* yeah...

Almighty-Alpaca:
The only thing you'd need for this is a database with a really fast access. Not sure what FredBoat currently uses, but redis for example could work.

Frederikam:
We use Postgres

Frederikam:
We actually cache our entities with ehcache, so it wouldn't really cause much delay, even if the database is located on a different continent

PR #440, opened 2018-01-23
[WIP] First pass on adding custom aliases

napstr:
Started off feeling like writing a nice little feature.
Parsing and entity stuff was trivial enough, and it works perfectly fine, but my head is spinning from trying to give it a decent UX.

- [ ] ~~merge after #436 and #439~~ / check migration versioning
- [ ] resolve dohz //todos
- [ ] (probably) write a docs page

### DIFF
--- a/FredBoat/src/main/java/fredboat/command/util/AliasCommand.java
+++ b/FredBoat/src/main/java/fredboat/command/util/AliasCommand.java
@@ -1,0 +1,199 @@
+/*
+ *
+ * MIT License
+ *
+ * Copyright (c) 2017 Frederik Ar. Mikkelsen
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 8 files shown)


napstr:
related to #168

napstr:
Please have a look:

Using an alias to have FredBoat say something:
image
Adding an essential alias which the bot creators failed to add:
image
Deleting an alias:
image
Aliases can be edited for a entire guild saying `guild` or `here` in the command:
image

napstr:
Some Questions:
1. Should aliases overwrite official FredBoat commands?
I think they should not. While it's funny to do so for some commands, we definitely have commands that we would not want to be overwritten.

2. If a guild and a user have defined the same alias, which one should FredBoat respond to?
I would think the guild alias takes precedence over an alias a user has defined for themselves.

napstr:
I'll take the lack of discussion around this to indicate that my I am free to answer my questions with whatever I prefer, and that the design is fine as it is. I shall bring this PR back up to speed with the current code and merge it then.

Frederikam:
>If a guild and a user have defined the same alias, which one should FredBoat respond to?
I would think the guild alias takes precedence over an alias a user has defined for themselves.

Why? A user's alias would be more specific. Furthermore, the user might not be aware of what aliases a guild is using

Frederikam:
It would be inconvenient and probably also rude if some user was to operate the bot in Danish when the accepted language of a guild might be English.

However, I think this isn't such a big a deal since it is unlikely that we are going to have many conflicts. Perhaps it is even confusing that both users and guilds have aliases.

napstr:
Someone else feel free to pick this up.

PR #449, opened 2018-02-02
Adjust hardcoding for FredBoat Hangout

Kurozume:
Adjust CommandManager.java so that the hardcoding for FredBoat Hangout now blocks all types of command (where it was previously just music commands) in channels that aren't #spam_and_music or #staff unless the user is Fre_d or Cynth


knight-ryu12:
I'd say check with PermSystem (as the Fre_d said), I think we gonna have hard time if we disabled entire thing

napstr:
Agree, admin check would be dope. And if the check isn't passed, I'd love an automatic answer to the user telling them to a) read the rules in #info_and_rules and b) link a channel where to use commands (#spam_and_music in our case), deleting both the message and FredBoat's response after a few seconds.

napstr:
Is this check necessary? I don't see the issue with enabling it on every bot (patron, CE).
` && guild.getJDA().getSelfUser().getIdLong() == BotConstants.MUSIC_BOT_ID`

There is also `DiscordUtils.isOfficialBot` for a wider bot id check.

PR #450, opened 2018-02-03
Add shorthand values to the SetAvatarCommand

Shredder121:
Also adds support for branding: urls, so we can use GitHub as CDN

I figured it wasn't necessary to put the main boat's avatars in there since it won't listen to the command 😂


Shredder121:
This makes avatar administration easier, when we need to swap them, and makes it easier to change branding when you selfhost
*Shrug*

napstr:
Looks good to me. Can you please a add a line somewhere about how to use the shortcut to the logos, preferably in the help?

PR #452, opened 2018-02-05
Implement repository pattern

napstr:
1k new lines for a rework (its more of an abstraction), OOP is magnificient.

Feel free to chip in, or don't.
This is happening in preparation of swapping out SqlSauce repositories for repositories that fetch entities from a REST backend, which we will need for the dashboard.
We do not want having several apps (dashboard backend, fredboat) connect to the database directly via JDBC, because that will make migrations a bit of a nightmare - they are already annoying enough with two independent bots connecting directly to the db using the same tables.


FredBoat/src/main/java/fredboat/db/EntityIO.java
```diff
@@ -115,86 +103,138 @@
...
+            operation.run();
+        } catch (DatabaseException e) {
+            log.error("EntityIO database operation failed", e);
+            throw new DatabaseNotReadyException(e);
         }
     }
 
 
-    // Loading of Entities
+    // Blacklist stuff
```
Shredder121: Don't you think when writing these comments we're combining a lot of things in one class?
[Just a thought](https://en.wikipedia.org/wiki/Interface_segregation_principle)

napstr: I do but I didn't really come up with a nice way to split it up ¯\_(ツ)_/¯

napstr:
Oh, looks like half of the lines added are just copyright notices of interfaces :^)

Frederikam:
Can you give a quick summary of what this does? I'm not in the mood to read 1500 code changes, even if much of it is licenses :stuck_out_tongue:

Shredder121:
It separates the code handling CRUD operations on entities into separate interfaces+impl
Although I, myself, would change the repo interface, and the abstract implementation to be generic, so it's not so copy-pasty.

FredBoat/src/main/java/fredboat/db/EntityIO.java
napstr: Thanks for the link. I split it up into interface and implemented them all in EntityIO. I think I can see the bigger picture, but I'm not sure if I did everything correctly / there are things missing.

FredBoat/src/main/java/fredboat/db/repositories/impl/SqlSauceGuildConfigRepo.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Nullable
+    @Override
+    public GuildConfig get(String id) {
+        return dbWrapper.getEntity(EntityKey.of(id, GuildConfig.class));
+    }
+
+    @Override
+    public void delete(String id) {
+        dbWrapper.deleteEntity(EntityKey.of(id, GuildPermissions.class));
```
Shredder121: This is what I mean with copy-pasty.

napstr:
Ye I just fixed that too :P

FredBoat/src/main/java/fredboat/db/EntityIO.java
Shredder121: Basically what I was thinking is a little more like spring-social-* projects([example-1][] [example-2][]
Where there is a '*Template' and *Operations'
so `thingTemplate.getOtherOperations().doThing()`
(Basically `EntityIO.getThingRepo().get(id)` in our case)
This doesn't pollute EntityIO with unrelated methods and segregates completion results/responsibilities, etc.

I'm not saying it has to be done that way, but it's one which I find rather elegant.
There is still the namespacing, but not the hassle to get an `IRepo<SpecificTypeNeeded>`.

[example-1]: https://github.com/spring-projects/spring-social-facebook/blob/master/spring-social-facebook/src/test/java/org/springframework/social/facebook/api/FeedTemplateTest.java#L243
[example-2]: https://github.com/spring-projects/spring-social-facebook/blob/master/spring-social-facebook/src/test/java/org/springframework/social/facebook/api/FeedTemplateTest.java#L256

napstr:
Thanks again, this review was great. Foolish me got slightly less foolish.

PR #469, opened 2018-02-17
Rest db backend

napstr:
This PR pulls out the FredBoat database into its own spring boot app.
This is implemented in an optional way, by sharing entity classes, meaning it is backwards compatible. This might change in the future (probability: likely).

Gson is used to serialize and deserialize entities from/to json. The reason for this choice is the zero annotations / configuration approach of it, as well as [superior benchmark values](https://blog.takipi.com/the-ultimate-json-library-json-simple-vs-gson-vs-jackson-vs-json/) for small json payloads (all our payloads are small according to the definition of that benchmark, and are going to stay small for the future).

Questionable practices of this PR (that I am kind aware of): 
- This is my first time building something with Spring boot
- Since some of our entity ids can be complicated, this is using POST requests for almost everything (instead of GET, DELETE, etc), because POST requests can have a body, where an entity id can be put into
- Controllers and Repos share the same interface

ToDo:
- [ ] What happens if migrations are applied to be backend, changing the look of an existing entity, with FredBoat running? Is that acceptable?
- [x] IO error handling in RestRepos
- [ ] Error/success communications layer
- [ ] API Documentation
- [x] replace client side caching


napstr:
One thing I forgot to mention is that this is also the thing that shall serve data for the dashboard, which will require some kind of authentication.
Small steps tho. Having a working, internal db rest server should help with developing and identifying issues.

Frederikam:
>One thing I forgot to mention is that this is also the thing that shall serve data for the dashboard, which will require some kind of authentication.

Why not just a token configured on both ends?

napstr:
"both ends" - one of them being untrusted client code? This requires further explanation.

My own ideas include implementing a DiscordOAuth solution, and setting a signed and optionally encrypted JWT token as a cookie to identify users and ensure the identification is genuine.

Frederikam:
>"both ends" - one of them being untrusted client code? This requires further explanation. My own ideas include implementing a DiscordOAuth solution, and setting a signed and optionally encrypted JWT token as a cookie to identify users and ensure the identification is genuine.

I thought you were talking about FredBoat<->Backend. Of course, we'd want to use OAuth with our own (not OAuth) tokens stored on the client.

napstr:
Sure, tokens to identify our own clients sounds good. We'll have to figure out a deployment involving one of private network / ssl / ssh tunnel.

I've also added another todo which is about handling caching. Using the backend moves the 2nd level cache away, a client side caching solution is required to replace that. I guess a simple guava cache per entities with values comparable to those of the current ehcache.xml files will do.

napstr:
The essential todos have been ticked off, this change is backwards compatible (backend will not be used by the bot unless a url is provided), anything else can be worked on in the dev branch.

PR #474, opened 2018-02-28
Remove duplicated backticks around prefix in prefix command

smileytechguy:


napstr:
Why though? Double backticks are rendered the same way single backticks are - with the exception that if the string they surround may contain (single) backticks without breaking. If anything, all our backticks should be double ones.

smileytechguy:
Really? Well, TIL.  I did not know that double backticks worked, I figured that they would be closing themselves so none of the actual text got formatted. Closing.

PR #482, opened 2018-03-12
Make volume persisted

Shredder121:
When the player is discarded, either by ;;destroy, or a fresh start of the JVM, the volume will always be set to default (100) regardless of what it should have been set to.

### DIFF
--- a/FredBoat/src/main/java/fredboat/command/music/control/VolumeCommand.java
+++ b/FredBoat/src/main/java/fredboat/command/music/control/VolumeCommand.java
@@ -53,17 +53,22 @@ public class VolumeCommand extends Command implements IMusicCommand, ICommandRes
         if (Launcher.getBotController().getAppConfig().getDistribution().volumeSupported()) {
 
             GuildPlayer player = Launcher.getBotController().getPlayerRegistry().getOrCreate(context.guild);
+            int oldVolume = (int) Math.floor(player.getVolume() * 100);
             try {
                 float volume = Float.parseFloat(context.args[0]) / 100;
                 volume = Math.max(0, Math.min(1.5f, volume));
 
+                int newVolume = (int) Math.floor(volume * 100);
                 context.reply(context.i18nFormat("volumeSuccess",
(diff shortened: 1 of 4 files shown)


FredBoat/src/main/java/fredboat/command/music/control/VolumeCommand.java
```diff
@@ -53,17 +53,22 @@ public void onInvoke(@Nonnull CommandContext context) {
...
 
+                int newVolume = (int) Math.floor(volume * 100);
                 context.reply(context.i18nFormat("volumeSuccess",
-                        Math.floor(player.getVolume() * 100), Math.floor(volume * 100)));
+                        oldVolume, newVolume));
 
                 player.setVolume(volume);
+                Launcher.getBotController().getEntityIO().transformGuildConfig(context.getGuild(),
+                        config -> config.setVolume(newVolume)
+                );
```
Shredder121: @napstr Before this change I had a hibernate validator annotation on the field, but figured it would be overkill since it requires this call to go first, handling of the ConstraintViolationExceptions, etc etc.
But what do you think of this change?

*GitHub didn't copy this in the diff.*

Shredder121:
Do you think shuffle and repeat would be nice to persist?

napstr:
Yes I think they should also be persisted, but that can be added later, no need to cram everything into one PR.
Just want to decide on the question of having the player values in their own entity or not.

Frederikam:
I agree with Napster. We should probably keep volume, shuffle and repeat in the same entity as the tracks

napstr:
A guild player endpoint has been added to quarterdeck: https://koakuma.fredboat.com/qd/swagger-ui.html#/guild-player-controller

Shredder121:
This one is superseded now? Do we call the right client setters?
Then this can be closed.

napstr:
It has not been implemented client side yet.

Frederikam:
Closing this as the player persistence I am working on should make this obsolete.

PR #490, opened 2018-04-11
Rework/moderation comms

napstr:
Notable changes:
- Better messages when something goes wrong
- Finer grained checks before doing anything
- Hackbans (bans by ids of users that the bot does not have in its cache)
- Unban command (by userId)
- Softban only requires the invoker to only have kick + manage messages
  permissions instead of ban permissions
- Message deletion length for ban commands configurable
- CompletionStage because sometimes costly lookups need to be done

Many thanks for testing to Praetor

Outstanding issues:
Very big and ugly help.
Commands use flags for some things.
 (these things are not average-end-user friendly)
This should demonstrate it:
image

The commands do take decent default values for everything, so the concerns are probably irrelevant /shrug


Frederikam:
The help string seems too verbose/confusing to me. Maybe elaborate that the user can be a mention, id or name. Then just have `user` (like we have it with the permission command)

image

image
Maybe also provide aliasses like -r and -d

napstr:
-r and -d are already a thing. No idea how to display them nicely and convey their meaning at the same time without it being confusing. 

Can't copy the permissions command structure like that. You're missing one detail, which is that when a username with spaces in it is involved, the reason has to be given behind the flag. We can't reliably and usefully search for a user AND distinguish the reason otherwise.

An alternative solution that I think would be tremendously more user friendly would be doing these in several steps, like asking the invoker each of the details.
However we have nothing in place to handle that and I would prefer not to bloat this PR.

napstr:
In staff chat (ty Otter) we wrapped up the following things:

Handling users with space in their name is an edge case that should be ignored for the current, single command / non conversational approach. A fuzzy search only looks for current members of a discord, so a mention should be possible just as well in most cases.

The deletion flag is actually "too much choice" - the big difference is deleting (1-7), or not deleting (0).
Simpler approach: By default, delete the Discord default value (1 day), and allow to pass an optional flag `-k / --keep` for 0 days.

Doing the above things simplifies the help and parsing by _a lot_.

napstr:
Picture of the current state of this
image

Frederikam:
image
Maybe allow search by username?

Other than that the UX looks good :+1:

napstr:
>Maybe allow search by username?

A banned user is not part of the guild, we would have to implement a search throughout the user cache, and I'm a bit wary of doing that just for this one use case, given the upcoming changes.

Frederikam:
>we would have to implement a search throughout the user cache

Is that so? You can see the names of people who are banned, but not in a mutual guild.

napstr:
Oh, right...the ban list.

PR #491, opened 2018-04-21
Changed the docker compose file to the correct branch on the Quarterdeck repo.

jtagt:


napstr:
There is no "dev" channel.
The correct channel is right there in the file.

napstr:
Also I don't understand why you're changing one of the component to the dev-channel, but not the other. The file does not make any sense that way, there is no stable v3 release of FredBoat either.

jtagt:
I see. I close this PR since there was no issue with the docket-compose file.

napstr:
If you need access to a backend in order to write client code, talk to me on Discord.

PR #497, opened 2018-05-13
Update docker-compose to match updated guide

shikhir-arora:
As https://github.com/FredBoat/fredboat.com/pull/28 is now live and people are using it, these should be changed over - especially Quarterdeck - since there's no `stable-v1` build (and never was 😆)

Since everyone following the updated guide will be aware of the selfhosting changes, and all pulling dev, I think the second change (to `dev-v3`) is fine as well. I've been running it for months.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -45,8 +45,8 @@ services:
     # FredBoat selfhosting website and/or the selfhosters channel (see top of this file for how to get to these places)
     # Once you use the dev branch, you may not be able to go back to stable without deleting your database.
     # To run on (some) arm based machines, prepend the tag with arm64v8-, for example: fredboat/quarterdeck:arm64v8-dev-v1
-    image: fredboat/quarterdeck:stable-v1
-    #image: fredboat/quarterdeck:dev-v1
+    #image: fredboat/quarterdeck:stable-v1
+    image: fredboat/quarterdeck:dev-v1
     restart: always
     labels:
       - "com.centurylinklabs.watchtower.enable=true"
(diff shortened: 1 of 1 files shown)


Frederikam:
I really don't want to move people over to dev. There's a reason we have a stable and a dev branch

shikhir-arora:
No worries, the QD was the most important, since there's no stable-v1. Changed back.

shikhir-arora:
If you prefer I can just remove the line referencing stable-v1 for Quarterdeck for now; it doesn't/hasn't existed.

Frederikam:
It is likely an error that quarterdeck stable-v1 doesn't exist. It has a master branch

Frederikam:
I don't think we want to mismatch the versions

shikhir-arora:
@napstr set it up that way, from the start, he's aware of it (even the tag on the Backend repo is dev-v1) - it's confusing at first because you'll get a docker error pulling stable-v1.

Frederikam:
I'd like him to clarify before merging this

shikhir-arora:
That's fine, I only remembered this as some users posted in #selfhosters and follows the guide - their only problem was the error thrown pulling a nonexistent docker image. Anyways, I know Napster said "..there may never be a stable Quarterdeck" before. Will wait on clarification.

Frederikam:
image
Works for me. The mismatch might not be a problem if we increment the major version with each breaking schema change.

napstr:
We should probably just promote anything that managed to run for 1-2 days on FredBoat to stable. There are no other criteria for something being stable afaik.

Frederikam:
I'm looking forward to deploying Sentinel. Then we'll be able to run tests and deploy frequently.

PR #502, opened 2018-06-13
Add spring application context to eval command

napstr:
Allows for usage in the eval command like:

```
;;eval return spring.getBean("shardManager").getGuildById("foo").getName()
```

Transforming the `Supplier<ApplicationContext>` through a kotlin -> java -> kotlin was a bit awkward, hope the code is correct.

### DIFF
--- a/FredBoat/src/main/java/fredboat/main/Launcher.kt
+++ b/FredBoat/src/main/java/fredboat/main/Launcher.kt
@@ -29,10 +29,13 @@ import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration
 import org.springframework.boot.autoconfigure.jdbc.DataSourceTransactionManagerAutoConfiguration
 import org.springframework.boot.autoconfigure.orm.jpa.HibernateJpaAutoConfiguration
 import org.springframework.boot.context.event.ApplicationFailedEvent
+import org.springframework.context.ApplicationContext
+import org.springframework.context.ApplicationContextAware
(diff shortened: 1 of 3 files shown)


Frederikam:
Nice example bean :stuck_out_tongue:

FredBoat/src/main/java/fredboat/main/Launcher.kt
```diff
@@ -179,6 +185,10 @@ open class Launcher(
         //        .anyMatch(shard -> shard.getStatus() != JDA.Status.CONNECTED);
     }
 
+    override fun setApplicationContext(applicationContext: ApplicationContext?) {
```
Frederikam: This probably shouldn't be nullable

FredBoat/src/main/java/fredboat/main/Launcher.kt
```diff
@@ -59,7 +62,9 @@ open class Launcher(
         private val invalidationAgent: GuildCacheInvalidationAgent,
         private val voiceChannelCleanupAgent: VoiceChannelCleanupAgent,
         private val carbonitexAgent: CarbonitexAgent
-) : ApplicationRunner {
+) : ApplicationRunner, ApplicationContextAware {
+
+    var springContext: ApplicationContext? = null
```
Frederikam: This should be `lateinit var` and nonnull

napstr:
Applied the requested changes.

Issue #495, opened 2018-05-08
;;weather recognition update


TernundaHunter:
![image](https://user-images.githubusercontent.com/31316028/39734138-7dffc5ea-523b-11e8-96a0-cc0cf8f36844.png)

![image](https://user-images.githubusercontent.com/31316028/39734149-8613279a-523b-11e8-9b8b-561fb91c60f3.png)

Since it does recognize Las Vegas and Chula Vista, I no longer assume it is simply incapable of reading strings.

PR #503, opened 2018-06-20
Weather command not parsing location names with spaces

SuhanKoh:
Fixing issue reported by user, #495

### DIFF
--- a/FredBoat/src/main/java/fredboat/command/util/WeatherCommand.java
+++ b/FredBoat/src/main/java/fredboat/command/util/WeatherCommand.java
@@ -41,7 +41,7 @@ public class WeatherCommand extends Command implements IUtilCommand {
         }
 
         String query = context.rawArgs;
-        String alphabeticalQuery = query.replaceAll("[^A-Za-z]", "");
+        String alphabeticalQuery = query.replaceAll("[^A-Za-z\\s]", "");
 
         if (alphabeticalQuery == null || alphabeticalQuery.length() == 0) {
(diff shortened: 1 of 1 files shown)


Frederikam:
The regex still won't work for town names containing characters that are not A-Z, for instance:
```
Tjæreborg
São Paulo
Москва
Clermond-Ferrand
東京
```
You might also need to `trim()` the string as well

Frederikam:
You might also want to target the `sentinel` branch, as this file has been migrated to Kotlin

SuhanKoh:
@Frederikam I feel like migration to kotlin for this change can be on a separate pr. I created this pr intended to fix the issue with spaces in the query

SuhanKoh:
I don't think openweather support searching 東京

Frederikam:
Perhaps not, but query parameters needs to be URL encoded. I don't know if the URL builder handles that

SuhanKoh:
@Frederikam It does, https://square.github.io/okhttp/3.x/okhttp/okhttp3/HttpUrl.Builder.html#addQueryParameter-java.lang.String-java.lang.String-

napstr:
Is it just me or does this look like a good case for a unit test?

Frederikam:
The API does not seem to support 東京 or Clermond-Ferrand, but the others work when using curl

Frederikam:
If I were you I would migrate to sentinel. I have developed an API to run integration tests on commands that you can use

PR #507, opened 2018-07-11
Re-add moderation commands

Frederikam:
This code is rather convoluted and prone to exploit (hence the need for review). I might write some integration tests, but there are a lot of edge cases.


FredBoat/src/main/java/fredboat/command/moderation/SoftbanCommand.kt
```diff
@@ -0,0 +1,178 @@
...
+        //A softban is like a kick + clear messages, so check for those on the invoker
+        if (!context.checkInvokerPermissionsWithFeedback(Permission.KICK_MEMBERS + Permission.MESSAGE_MANAGE)) return false
+        //We do however need ban perms to do this
+        if (!context.checkSelfPermissionsWithFeedback(Permission.BAN_MEMBERS)) return false
+
+        // if the target user is NOT member of the guild AND banned, the invoker needs to have BAN perms, because it is
+        // essentially an UNBAN, and we dont want to allow users to unban anyone with just kick perms
+        if (targetMember == null) {
+            var isUnban = false
+            fetchBanlist(context)!!.doOnNext {
```
napstr: This _could_ be a source of NPEs due to race conditions of the permissions check? I think it's also present in the original code.
Maybe the fetchBanlist method can be written in a way to not return null?

Frederikam: That always felt weird to me, but this should only cause a NPE in ultra-rare cases

Frederikam:
The commands that have not been thoroughly integration tested have been disabled. This should be safe to merge.

PR #515, opened 2018-07-20
Update CONTRIBUTING.md

Frederikam:
This PR updates our contributing guidelines so that it stays more up to date with changes related to Sentinel and integration testing. I still seek a way to tackle how we can make it easy to run the necessary services (quarterdeck, rabbitmq, sentinel, etc) for development.

*Accidentally branched off of the dev branch, but we should be fine pulling those other two commits.*


Frederikam:
Moved to a different branch with config changes

PR #525, opened 2018-09-20
bumped JDA to 3.8.0_425

SixAiy:
i have tested all the files and rebuilt the project and everything seems to work fine as its currently set so i suggested a JDA bump to 3.8.0_425 for newest version of JDA


Frederikam:
Sorry, I didn't notice this PR. This change has already been made, though this change alone would have caused it to fail due to breaking changes in JDA.

PR #526, opened 2018-09-30
Replace akinator command with link to aki bot

napstr:
as discussed in chat, the "permanent" solution


Frederikam:
You may also want to contribute this to the sentinel branch as a `TextCommand`

Issue #520, opened 2018-07-29
make ;;uptime only show uptime and not additional stats with it!


thisisjvgrace:
So, when we execute `;;uptime` command, FredBoat also sends the additional stats which kind of looks like spam. It's like, you execute `;;np` command but FredBoat would also posts the queue output along with now playing info. :D (Speaking from a selfhoster's perspective)

otherwise, FredBoat kun is very dank bot. 

![2018-07-29_16-26-03](https://user-images.githubusercontent.com/24418520/43365564-5e1a2cb4-934c-11e8-99ae-b703dcca9558.png)

With all due respect,
Jyu Viole Grace.

smriga-jan:
Hello, 
I can see that command ";;uptime" is only an alias for ";;stats command". 
Can I create a separate command ";;uptime" to only show uptime?

Frederikam:
@smriga-jan sorry for the late reply.

You're welcome to, although you'll want to work on the `sentinel` branch. Please read its contribution guide: https://github.com/Frederikam/FredBoat/blob/sentinel/CONTRIBUTING.md

smriga-jan:
@Frederikam thank you for your reply!

I will read this guide carefully and look into it.

PR #537, opened 2018-11-05
Add uptime command to only show uptime based on issue #520.

smriga-jan:
Opened a pull request with solution to issue #520.

Created new separate command to show uptime only, registered it in CommandInitializer and also added new paragraph in languages for translations.

Hopefully everything is done correctly.


Frederikam:
Sorry for the late reply. It seems my email service marked this PR's notifications as spam.

It seems like this PR is based upon `master`, but is aimed at `sentinel`. These two branches are very different. This means that the command you wrote will need to be rewritten for Sentinel, in the manner explained in the contribution guide. This PR would otherwise only be compatible with legacy FredBoat.

Sentinel will likely be merged into dev and master within a couple days. I apologize for the confusion.

PR #538, opened 2018-11-21
Add uptime command

smriga-jan:
Rewritten ;;uptime command to be compatible with Sentinel branch.

### DIFF
--- a/FredBoat/src/main/java/fredboat/command/info/UptimeCommand.kt
+++ b/FredBoat/src/main/java/fredboat/command/info/UptimeCommand.kt
@@ -1,0 +1,84 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2017 Frederik Ar. Mikkelsen
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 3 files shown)


FredBoat/src/main/java/fredboat/command/info/UptimeCommand.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    override suspend fun invoke(context: CommandContext) {
+        context.reply(getUptime(context))
+    }
+
+    override fun help(context: Context): String {
+        return "{0}{1}\n#Shows uptime of this bot."
+    }
+
+    companion object {
```
Frederikam: Why are you using a companion object? `getUptime` is only used from a non-static context. Without it, the `botId` variable can be removed as well.

FredBoat/src/main/java/fredboat/command/info/UptimeCommand.kt
```diff
@@ -0,0 +1,84 @@
...
+    init {
+        botId = selfUser.id
+    }
+
+    override suspend fun invoke(context: CommandContext) {
+        context.reply(getUptime(context))
+    }
+
+    override fun help(context: Context): String {
+        return "{0}{1}\n#Shows uptime of this bot."
```
Frederikam: This should also be translated

Frederikam:
This was resolved by https://github.com/Frederikam/FredBoat/pull/557

PR #553, opened 2019-01-15
PlayNext Command

Nanabell:
Duno. was bored. did this.
I think there was an issue for this but I can't find it anymore

Dunno if this design is acceptable this is just the first idea that came to my mind.
I simply reused the PlayCommand with a boolean to indicate that its a topQueue request.
Also Playnow will require DJ permission since it directly modifies the queue.

- [x] still, need to add help text for PlayNow command


FredBoat/src/main/java/fredboat/commandmeta/CommandInitializer.kt
```diff
@@ -235,6 +235,9 @@ class CommandInitializer(cacheMetrics: CacheMetricsCollector, weather: Weather,
         musicModule.registerCommand(PlayCommand(playerLimiter, trackSearcher, videoSelectionCache,
                 listOf(SearchProvider.SOUNDCLOUD),
                 SOUNDCLOUD_COMM_NAME, "sc"))
+        musicModule.registerCommand(PlayCommand(playerLimiter, trackSearcher, videoSelectionCache,
+                listOf(SearchProvider.YOUTUBE, SearchProvider.SOUNDCLOUD),
+                "playnext", "pn", topQueue = true))
```
Frederikam: Alias `playtop` would be nice.

Nanabell:
Did some tests regarding playnext playlists & playnext with shuffle enabled
- playnext with playlist does not work and I don't think its necessary to
- playnext with shuffle does not currently work, I was going to implement a way around this but after second thoughts would this even be a wanted behavior? I would assume if I shuffle a playlist I want it shuffled. The other option that I would see is to just disable the playnext command if the queue is shuffled saying that you need to unshuffle.

Frederikam:
I believe that the DJ's deliberate commands should be followed. If the DJ wants to add an arbitrary number of tracks to be played next, then so be it. Shuffle mode is also handled by the DJs, so if a DJ wants to bypass that, then so be it.

More precisely:
* Adding a playlist should be a legit use-case. A playlist might even be just a couple tracks after all.
* Adding a playlist while shuffling should not affect the order that the tracks of the playlist is played
* If the queue is shuffled by either enabling shuffle or invoking `;;reshuffle`, then the tracks added beforehand should be treated like any other.

Nanabell:
- [x] Adding a playlist with playnext works
- [x] unless shuffle will be enabled(while priority tracks already in the queue) or reshuffle tracks queued with playnext will always stay at the top
- [x] shuffle or reshuffle command will transform prioroty tracks into normal tracks allowing them to be shuffled

FredBoat/src/main/resources/lang/en_US.properties
```diff
@@ -1,5 +1,5 @@
 #X-Generator: crowdin.com
-playQueueEmpty=The player is not currently playing anything. Use the following syntax to add a song\:\n;;play <url-or-search-terms>
+playQueueEmpty=The player is not currently playing anything. Use the following syntax to add a song\:\n{0}{1} <url-or-search-terms>
```
Frederikam: This change looks harmless, but it would actually require invalidating all of our current translations. It's also rather tedious to update the source strings on Crowdin.

Nanabell: That is true but is it really a good idea to have command name and prefix hardcoded into the strings?

Frederikam: No, although it comes to mind that you can just do something like `"...;;play...".replace(";;", prefix)`

PR #557, opened 2019-01-30
Fix #520 - Separate Uptime and Stats commands.

### LINKED ISSUE #520: make ;;uptime only show uptime and not additional stats with it!

OtterBoops:
Also stripped some of the imports from Stats, since they seem unnecessary.

### DIFF
--- a/FredBoat/src/main/java/fredboat/command/info/UptimeCommand.kt
+++ b/FredBoat/src/main/java/fredboat/command/info/UptimeCommand.kt
@@ -1,0 +1,58 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2017 Frederik Ar. Mikkelsen
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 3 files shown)


Frederikam:
I wonder if this really warrants translation. It is more of a maintenance command, albeit available to the public.

OtterBoops:
> I wonder if this really warrants translation. It is more of a maintenance command, albeit available to the public.

Alternative would be to simply have `Online for/Uptime: x Days, y Hours, z Minutes, a Seconds`

Frederikam:
I think that would be much easier, albeit you are using wrong capitalization in your comment.

FredBoat/src/main/java/fredboat/command/info/UptimeCommand.kt
```diff
@@ -51,7 +51,7 @@ class UptimeCommand(name: String, vararg aliases: String) : Command(name, *alias
         val mins = (totalSecs / 60 % 60).toInt()
         val secs = (totalSecs % 60).toInt()
 
-        val response = context.i18nFormat(("uptimeParagraph"), days, hours, mins, secs)
+        val response = MessageFormat.format(("Online for: {0} days, {1} hours, {2} minutes and {3} seconds"), days, hours, mins, secs)
```
Frederikam: Please use Kotlin templates instead of MessageFormat. E.g `"Online for $days days"`

You should also remove the extra pair of parenthesis

Issue #563, opened 2019-02-15
username shows nickname


eiiiiix:
i was using `;;nowplaying` to check how much time was left in a song and noticed that it shows the nickname then discriminator

not really sure if this is an issue but it bothers me, most bots show username instead of nickname

![image](https://user-images.githubusercontent.com/30561378/52827778-2543bc80-3094-11e9-96fc-f9cb9e7b9e7f.png)

Frederikam:
Perhaps the problem is more that it shows the discrim. It seems appropriate to show the nickname instead of the username.

mcbobby123:
@Frederikam you mentioned in the other thread that this was fixed, but it isn't. Is this behavior intended? Seems confusing to me.

PR #565, opened 2019-02-15
Rewrite `pause` command in Kotlin.

OtterBoops:


Frederikam:
Magically teleporting this to the 5.0 branch due to how i18n works 

(PS: Don't tell GitHub that this fixes #563 and closes #564)

Frederikam:
Turns out I already migrated that file months ago. I'll merge the other PR for the i18n fix.

PR #566, opened 2019-02-23
Feature/player limitations

### LINKED ISSUE #39: Command to Limit the Queue count & Max Video lenght
The idea seems quite good. 
i duno if you want to wait for the dashboard to be done with this since its on the lower end of the priority  list

### LINKED ISSUE #357: Proposal for "Max Song Length"
As we already have a Staff managed "Max song length" (not to confuse with "max queue length") policy in Fredboat Hangout it would be nice if this was automated. I was thinking I bit about how to implement this the best way. The most obvious way to implement this would be to prevent users from queueing a song if it is above a set limit (e.g 10min).
While it would work perfectly fine what about users who are alone or if a group of users **want** to listen to a 1h remix. My proposal would be

Make tracks which exceed the max song length skippable by `USER` permission. This way everyone can queue songs with whatever length they want but we would prevent users from abusing long songs to prevent other users from playing songs. There is still something to be said about the possible abuse that anyone could join and skip the current track like this, but that's why I'm opening this issue. What do you think of this idea? Should we consider this or just stick to the "default" way of preventing long songs altogether?

### LINKED ISSUE #365: User queue cap
Have an option in ;;config where individual queue limits can be set. Recommend set value options like 5, 10, 20, so on. This is a good spam prevention tool, as well.

Nanabell:
Implement `QueueLimiter` (all naming is TBD and not final) which can take any number of `QueueLimits` to apply a number of limitations while queueing a track/tracks.

- 2 new methods in `GuildPlayer` `queueLimited(AudioTrackContext)` `queueLimited(List<AudioPlaylistContext>)`
- Add `AudioPlaylistContext` to identify a TrackContext being a playlist or Single track (inherrits from `AudioTrackContext`
- Added new "Base" `NullableContext` which acts as base for the `Context` but with nullable abstract properties to allow AudioTrackContext to extend it
- Added `queuePrevented` Metric
- Added 4 queue-limits (NoPlaylist, MaxTrackLength, MaxTrackCount, MaxUserTrackCount)

closes #39 
closes #357 
and closes #365


FredBoat/src/main/java/fredboat/audio/player/guildPlayerUtils.kt
```diff
@@ -69,6 +69,12 @@ val GuildPlayer.trackCountInHistory: Int
 val GuildPlayer.isHistoryQueueEmpty: Boolean
     get() = historyQueue.isEmpty()
 
+fun GuildPlayer.userTackCount(userId: Long): Int {
```
Frederikam: Should be `getUserTrackCount`

FredBoat/src/main/java/fredboat/audio/queue/ITrackProvider.kt
```diff
@@ -70,28 +70,19 @@ interface ITrackProvider {
...
-     * @param track add a track to the queue
-     */
-    fun add(track: AudioTrackContext)
-
-    /**
-     * @param tracks add several tracks to the queue
+     * @return amount of tracks in the queue filtered by userId
      */
-    fun addAll(tracks: Collection<AudioTrackContext>)
+    fun size(userId: Long): Int
```
Frederikam: `getCountByUser(...)` makes more sense here

FredBoat/src/main/resources/lang/en_US.properties
```diff
@@ -91,6 +91,11 @@ loadPlaylistTooMany=Added {0} tracks. Found too many tracks to display.
 loadErrorCommon=Error occurred when loading info for `{0}`\:\n{1}
 loadErrorSusp=Suspicious error when loading info for `{0}`.
 loadQueueTrackLimit=You can't add tracks to a queue with more than {0} tracks! This is to prevent abuse.
+loadPlaylistDisabled=TODO loadPlaylistDisabled
+loadMaxTracksExceeded=TODO loadMaxTracksExceeded
+loadMaxUserTracksExceeded=TODO loadMaxUserTracksExceeded
+loadMaxTrackLengthExceeded=TODO loadMaxTrackLengthExcceded
```
Frederikam: Nice strings

FredBoat/src/main/java/fredboat/messaging/internal/NullableContext.kt
```diff
@@ -0,0 +1,136 @@
...
+import fredboat.shared.constant.BotConstants
+import fredboat.util.TextUtils
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import reactor.core.publisher.Mono
+import java.text.MessageFormat
+import java.util.*
+import javax.annotation.CheckReturnValue
+
+abstract class NullableContext {
```
Frederikam: What is the point of this class? It looks bad practice use use `NullableContext` as a type rather than `CommandContext?`

FredBoat/src/main/resources/lang/en_US.properties
Nanabell: Suggest something. maybe @OtterBoops has an idea 😄

FredBoat/src/main/java/fredboat/messaging/internal/NullableContext.kt
Nanabell: I wanted to make use of the `Context` abstract class for the `AudioTrackContext` since both are a "context" but the AudioTrackContext can guarantee that the `textChannel` is not null https://github.com/Frederikam/FredBoat/blob/bfb51e0f967ace1a63be22e8a4a998f4b9455907/FredBoat/src/main/java/fredboat/audio/queue/AudioTrackContext.kt#L72-L77
since just duplicating the methods I need didn't to seem like a good idea either

Frederikam:
The problem with the new strings is that they don't imply that the limits are set by the server's admins

FredBoat/src/main/java/fredboat/command/config/ConfigCommand.kt
```diff
@@ -31,17 +31,52 @@ import fredboat.commandmeta.abs.CommandContext
...
         get() = PermissionLevel.BASE
 
+    private val configOptions: MutableList<ConfigOption> = mutableListOf()
+
+    init {
+        val booleanPredicate = Predicate<String> { it.equals("true", true) || it.equals("false", true) }
+        val nullableIntPredicate = Predicate<String> { (it.toIntOrNull() != null && it.toInt() > -1) || StringUtils.equalsAnyIgnoreCase(it, "none", "unlimited") }
+        val nullableLongPredicate = Predicate<String> { StringUtils.equalsAnyIgnoreCase(it, "none", "unlimited") || TextUtils.parseTimeString(it) != 0L }
+
+        configOptions.add(ConfigOption("track_announce", booleanPredicate,
```
Frederikam: These calls to `#add()` would be more readable if you put one arg on its own line

FredBoat/src/main/java/fredboat/command/config/ConfigCommand.kt
```diff
@@ -74,30 +110,39 @@ class ConfigCommand(name: String, private val repo: GuildSettingsRepository, var
...
 
     override fun help(context: Context): String {
         val usage = "{0}{1} OR {0}{1} <key> <value>\n#"
         return usage + context.i18n("helpConfigCommand")
     }
 }
+
+private data class ConfigOption(
+        val name: String,
+        val valueFormat: Predicate<String>,
```
Frederikam: Validator would be a more descriptive name. Also consider using a [typealias](https://kotlinlang.org/docs/reference/type-aliases.html):
```
typealias Validator = Predicate<String>
```

FredBoat/src/main/java/fredboat/command/config/ConfigCommand.kt
```diff
@@ -74,30 +110,39 @@ class ConfigCommand(name: String, private val repo: GuildSettingsRepository, var
...
+        val getter: (GuildSettings) -> String,
+        val setter: (GuildSettings, String) -> Unit
+) {
+    fun update(repo: GuildSettingsRepository, context: CommandContext, value: String) {
+        repo.fetch(context.guild.id)
+                .doOnSuccess { setter(it, value) }
+                .let { repo.update(it) }
+                .subscribe { context.replyWithName("`$name` ${context.i18nFormat("configSetTo", value)}") }
```
Frederikam: It would be better practice to return the Mono rather than subscribing to it here --- especially if we are going to do a web dashboard with settings and want to reuse this.

Frederikam:
BTW I really like what you did with the ;;config command, asides from those two comments.

Frederikam:
It is worth noting that one of the test cases is expecting the phrase "...will now play" where "...has been added to the queue" is given.

Nanabell:
> It is worth noting that one of the test cases is expecting the phrase "...will now play" where "...has been added to the queue" is given.

The test works if run alone 🤔. from what I've debugged til now it seems that the player state `isPlaying` is not reset from the test `search(PlayerRegistry)`

FredBoat/src/main/java/fredboat/command/config/ConfigCommand.kt
```diff
@@ -137,11 +140,10 @@ private data class ConfigOption(
         val getter: (GuildSettings) -> String,
         val setter: (GuildSettings, String) -> Unit
 ) {
-    fun update(repo: GuildSettingsRepository, context: CommandContext, value: String) {
-        repo.fetch(context.guild.id)
+    fun update(repo: GuildSettingsRepository, context: CommandContext, value: String): Mono<GuildSettings> {
```
Frederikam: This method can be shortened to an expression

Frederikam:
The only thing that needs to be fixed now is the failing test case. Try using the CI server to see which commit is at fault.

Nanabell:
> Does this not imply that `isPlaying` is broken?

No, but since its reactive `loader.player.queueLimited(atc)` will actually do the queueing and start playing the track and subscribe happens after that so in the subscribe part of the mono `isPlayling` will always be true

Frederikam:
I don't think we would have ever noticed that bug, had I not written some integration tests :smile:

PR #573, opened 2019-03-01
New persistence system

### LINKED ISSUE #139: Fix persistent queues
We have a persistent queue system, only that's it's broken and thus disabled.

I'm thinking of rewriting this, but with an SQL table instead of the filesystem. Thoughts?

Frederikam:
(Also a lot of other features that crept in)

Uses Spring Data for database access, and persists queue state to the database.

Fixes #139


.flooignore
```diff
@@ -0,0 +1,11 @@
...
+node_modules
+tmp
+vendor
+.idea/workspace.xml
+.idea/misc.xml
+.idea
+
+fredboat.yml
+fredboat.yaml
+commons.yml
```
napstr: What are these .floo files?

Frederikam: Used for live code sharing https://floobits.com/

PR #574, opened 2019-03-03
Feature/queue rework

Nanabell:
Replace the current Queue system with a more "modular" approach.
- IQueueHandler base interface handles all the basic Queue logic, like adding, removing, taking... tracks as well as to provide event style methods to implementations what happened when.
- SimpleQueueHandler: has no logic and was only used by me as logging debugger (usage TBD)
- ShufflableQueueHandler: implement shuffle logic into the queue system
- RoundRobinQueueHandler: allow the queue to provide tracks in a RoundRobin fashion (also called "Fair Queue" at times). Also when RoundRobin and Shuffle are enabled the Queue will provide Tracks still in RoundRobin order but shuffle the order of tracks per user
- RepeatableQueueHandler: adds repeat SINGLE & ALL capabilities to the queue

Remaining TODOS:

- [x] RoundRobin is currently Hardcoded enabled this needs to be saved to the GuildPlayer (including persistence) 
- [x] Command to toggle RoundRobin
- [x] All new classes are in the package named `tbd` well determinate that name now
- [ ] other things I forgot


FredBoat/src/main/resources/lang/en_US.properties
```diff
@@ -21,6 +21,9 @@ shuffleOn=The player is now shuffled.
 shuffleOff=The player is no longer shuffled.
 reshufflePlaylist=Queue reshuffled.
 reshufflePlayerNotShuffling=You must first turn on shuffle mode.
+roundRobinOn=The player is now in Round-Robin mode.
+roundRobinOff=The player is no longer in Round-Robin mode.
```
napstr: I feel a more end-user friendly name like for example 'fair mode' could be worth exploring.

Frederikam: It might also be a good idea to explain what it does when it is enabled

FredBoat/src/main/java/fredboat/command/music/control/RoundRobbinCommand.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    override suspend fun invoke(context: CommandContext) {
+        val player = getBotController().playerRegistry.awaitPlayer(context.guild)
+        player.isRoundRobin = !player.isRoundRobin
+
+        context.reply(if (player.isRoundRobin) "roundRobinOn" else "roundRobinOff")
+    }
+
+    override fun help(context: Context): String {
+        return "{0}{1}\n# " + context.i18n("helpRoundRobinCommand")
```
Frederikam: This key is missing

FredBoat/src/main/java/fredboat/commandmeta/CommandInitializer.kt
```diff
@@ -259,6 +259,7 @@ class CommandInitializer(
         musicModule.registerCommand(SelectCommand(videoSelectionCache, "select",
                 *buildNumericalSelectAliases("sel")))
         musicModule.registerCommand(ShuffleCommand("shuffle", "sh", "random"))
+        musicModule.registerCommand(RoundRobbinCommand("roundrobin", "rr"))
```
Frederikam: We might want to brand this as "fair" mode, in which case this would need to reflect that

FredBoat/src/main/resources/lang/en_US.properties
```diff
@@ -255,6 +255,7 @@ helpRepeatCommand=Toggle between repeat modes.
 helpReshuffleCommand=Reshuffle the current queue.
 helpSelectCommand=Select one of the offered tracks after a search to play.
 helpShuffleCommand=Toggle shuffle mode for the current queue.
+helpRoundRobinCommand=Toggle "Round Robin" / "Fair Queue" mode.\n#In this mode the queue will be sorted into rounds whereby it is guaranteed that every unique users can place once per round.
```
Frederikam: There are a few mistakes here:
`users` should be `user`
`place` should become `play music` or `play a track`
`whereby` could maybe be simplified to `where`, as out audience isn't necessarily so savvy at English.

Frederikam:
Sorry about the delay. I hadn't noticed the updates

PR #595, opened 2019-11-26
Add not recognized

Qlenglet:
This modification is made to avoid a bad command typing


Frederikam:
We had something like this years ago. I removed it because I got complaints of unintended invocations.

Best practice is to do nothing.

napstr:
Could be a guild settings that is off by default.

PR #599, opened 2019-12-09
Playlist command

kthezelais:
I add a playlist command. This feature is used to store tracks into a parallel list of the queue, and this list of tracks can be run on the queue when the user use the command ;;playlist run.

You can find some details on how to use this new command below :

The command ;;playlist <option> Allows you to configure the playlist:
    - add <search>: add a title from a search. Then enter the command ;;playlist [1-5] to choose a title from the search list.

    - copy: copy the queue into the playlist. If the playlist is not empty, enter the command ;;playlist [1-2] (1: overwrite the playlist, 2: do nothing).

    - list: display title's list of the playlist.

    - remove <option>: delete one or more titles from the playlist:
         - without option: displays the list of titles in the playlist.
         - all: delete all the titles in the playlist.
         - <number>: delete the title associated with the <number> parameter.

    - run <option>: adds all the titles from playlist to the queue.
         - without option: add the titles in order.
         - random: add the titles in the mess.


FredBoat/src/main/java/fredboat/command/music/control/PlayListCommand.java
```diff
@@ -0,0 +1,367 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static fredboat.main.LauncherKt.getBotController;
+import static fredboat.util.MessageBuilderKt.localMessageBuilder;
+
+public class PlayListCommand extends JCommand implements IMusicCommand, ICommandRestricted {
+
+    private List<AudioTrackContext> playlist;
```
Frederikam: This fundamentally doesn't work. You're sharing the same playlist between everyone. It also doesn't persist between restarts. 

Doing something like this this the right way would require working with the database, and even then I would do things differently.

PR #601, opened 2019-12-16
French Webradio functionality

superocket:
This contribution makes it possible to launch French web radio stations, using the HTTP protocol. The first step is to retrieve the URL of the webradio then launch them on the player. I had to modify the AudioPlayerManagerConfiguration file to integrate the http streams. Although these feeds must remain disabled, can you help me to activate them only when the webradio command is launched?


superocket:
This is how it works:

The main command is ;;webradio you have an alias ;;webr

```
;; webradio list : give the name of all available webradios

;; webradio *NRJ France* : play the webradio NRJ France for example

PR #608, opened 2020-03-25
Adjust logging for Sentinel

OtterBoops:
People don't like 90Gb log files


config/templates/docker-compose.yml
```diff
@@ -89,6 +89,10 @@ services:
     - "SPRING_RABBITMQ_ADDRESSES=amqp://guest:guest@rabbitmq"
     # Need a bigger memory size or any other custom JVM args? Uncomment and edit the line below accordingly
     #entrypoint: java -Xmx512m -jar sentinel.jar
+    logging:
+      options:
+        max-size: "1g"
+        max-file: "30"
```
Frederikam:
```suggestion
        max-file: "3"
```
30GB still seems a bit much

OtterBoops: Coffee isn't enough. I thought that was max-history. I'd probably say that max file is 1.

PR #621, opened 2020-06-10
Add command to play the first search result

### LINKED ISSUE #99: Skip the selection of tracks when searching
A command that plays the #1 searched track by default, or a provided number

example:
`;;instaplay <searchterm>`
and
`;;instaplay <[0-9]?> <searchterm>`

Maybe come up with a more clever/shorter name than instaplay, I just made that one up.

Picked this feature request up from the #suggestions channel

OtterBoops:
Fixes #99 

There are a bunch of stuff that can be discussed/changed:
- The name of the command
- A dedicated help entry
- A lot of cleanup
- Other misc requests


FredBoat/src/main/java/fredboat/command/music/control/PlayFirstCommand.kt
```diff
@@ -0,0 +1,192 @@
...
+
+
+    override fun help(context: Context): String {
+        val usage = "{0}{1} <url> OR {0}{1} <search-term>\n#"
+        return usage + context.i18nFormat(if (!isPriority) "helpPlayCommand" else "helpPlayNextCommand", BotConstants.DOCS_URL)
+    }
+
+    companion object {
+
+        private val log = LoggerFactory.getLogger(PlayCommand::class.java)
```
Frederikam, line 188:
```suggestion
        private val log = LoggerFactory.getLogger(PlayFirstCommand::class.java)
```

FredBoat/src/main/java/fredboat/command/music/control/PlayFirstCommand.kt
```diff
@@ -0,0 +1,192 @@
...
+
+            }
+        }
+    }
+
+
+
+    override fun help(context: Context): String {
+        val usage = "{0}{1} <url> OR {0}{1} <search-term>\n#"
+        return usage + context.i18nFormat(if (!isPriority) "helpPlayCommand" else "helpPlayNextCommand", BotConstants.DOCS_URL)
```
Frederikam, line 183: Bad strings

FredBoat/src/main/java/fredboat/command/music/control/PlayFirstCommand.kt
```diff
@@ -0,0 +1,192 @@
...
+
+                val track = list.tracks.first()
+                
+                val player = Launcher.botController.playerRegistry.getOrCreate(context.guild)
+                
+                val invoker = context.member
+
+                val outputMsgBuilder = StringBuilder()
+
+                //Technically, you always pick the first song. Kind of dirty.
```
Frederikam, line 164: That's probably fine. It hints at what this command is for

Frederikam:
This isn't a very clean implementation. It introduces a lot of duplicate code.

Maybe it would be better if we added this sort of functionality to `PlayCommand`? We could specify a flag like `-1` or `--first`

OtterBoops:
> This isn't a very clean implementation. It introduces a lot of duplicate code.
> 
> Maybe it would be better if we added this sort of functionality to `PlayCommand`? We could specify a flag like `-1` or `--first`

I really don't like the flag idea. In many cases, it may overlap with the actual song name in a way.
How about an alias to the command, similar to sentisneks?

Frederikam:
Which songs could it possibly overlap with?

OtterBoops:
Anything that first or the number 1 in the name. Ultimately it's up to you, but from my perspective, a separate alias/command is the way to go.

Frederikam:
You're ignoring the hyphens. We'd be searching for ` -1 ` not `1`

Nanabell:
to be honest from a ux perspective making it a flag instead of a separate command seems terrible,
will only cause it to not get used and flags are a bit cumbersome to use in discord
i'd be in favor of using a separate command

